Issue #14, opened 2019-12-23
As a user, i want to edit my personal details


sfinn85:

ryanlerch:
Duplicate of #42

PR #62, opened 2020-01-08
Coding standards

abompard:
This PR adds the files to do the required checks in CI (see #63).
It also runs black on the codebase in a separate commit that does not do anything else, so it's pure formatting.

An account on CI [has been requested](https://bugs.centos.org/view.php?id=16901) with ci.centos.org and I'm waiting for an answer from them to actually test CI.


abompard:
Actually, this is not related to #14 but to [freeipa-fas#14](https://github.com/fedora-infra/freeipa-fas/issues/14), we still need to move that issue to this project.

abompard:
OK I just copied the issue over, it's #63 here.

abompard:
Since we don't know how log it may take for the CentOS folks to create the project in Centos CI, and since that PR includes a black run that touches almost all files, I think it's still worthwhile to review / merge this PR now so it does not conflict with further work.

relrod:
As noted on IRC last week, I'm -1 on using Black. I'm particularly -1 on changing all `'` to `"` for no reason.

abompard:
I think Black brings a lot of benefits in a team of more than 1 person, because it brings the coding styles of people closer together, it makes it easier for someone to read anybody else's code and write according to the project's standards, and it makes PR reviews simpler to review because diffs will be smaller and more uniform. And all other advantages that the [black website](https://black.readthedocs.io/en/stable/index.html) lists.

However, I am willing to leave the strings as-is. I thought that this was the right time to integrate black because it is the start of the project *for me*, but I understand that it's a codebase that you have worked on for a while now and where you see a lot of history. I have underestimated that and I don't want to antagonize anyone.

I'm going to re-do the second commit with the option to skip string normalization. Would this address your concern @relrod?

sfinn85:
Hi @abompard & @relrod 
On our stand up on Thursday, we briefly discussed the pro's and con's to moving to black. There was a few pro's highlighted the main one being Coherence, we also briefly discussed the impact that would have on the code base. It was flagged that it would have little or no effect and so the decision was made to move to black " . Have we learnt more about it's impact v value leading to this decision being reverted ? 

Just to flag the decision made here, should reflect the facts regarding what is the best fit for the feature when released and in the future, not our personal preference's.

abompard:
Just to make things clear: this PR still integrates black, but in this version I have disable the string normalization feature that Rick is the most opposed to (if I understood correctly). The rest of the Black features are still on.

sfinn85:
OK thanks for clarifying @abompard . @relrod are you happy that this is the best decision for the overall quality of the feature delivery?

relrod:
@abompard Thanks. I'm more okay with it if the quotes are left alone (just because of how many lines would get changed). But I'm curious on some things.

1) Do you actually find

```python
     gpgkeys = FieldList(StringField('GPG Keys', validators=[Optional(),])) 
```

more readable than

```python
    gpgkeys = FieldList(
        StringField(
            'GPG Keys',
            validators=[
                Optional(),
            ]))
```
?

Maybe it's what I'm used to, but having it split up into multiple lines makes that *way* more parsable to me at a glance. Not to mention, adding or changing a validator is a one line diff here, but would very quickly hit the line length limit and turn into a multi-line diff using the "new" style.

2) I've never seen this style before, it feels odd:

```python
        flash(
            'You got it! %s has been removed from %s.' % (username, groupname), 'green'
        )
```

Usually if a closing paren is on a new line, it's because the function arguments are broken up into lines.

It's also inconsistent (if we're going for "coherence" - whatever is meant by that here). In the same file we have:

```python
        flash('You got it! %s has been added to %s.' % (username, groupname), 'green')
```
... with the closing paren on the same line

and earlier we have:

```python
    return render_template(
        'group.html',
        group=group,
        members=members,
        sponsors=sponsors,
        sponsor_form=sponsor_form,
        remove_form=remove_form,
        current_user_is_sponsor=current_user_is_sponsor,
    )
```

... where each argument gets its own line.

If we're going for "coherency", can we tell Black to....be coherent?

abompard:
Black is coherent within the limits of the number of characters per line that we tell it to use, that's why you see wraps for some lines and not for others. We can consider changing that limit, but I'm not sure that's useful.
If you wonder how the formatting is decided, they have [a detailed page](https://black.readthedocs.io/en/stable/the_black_code_style.html#how-black-wraps-lines) with reasons in their documentation.

What I mean by coherence is not only coherence thoughout the codebase (line wrapping limits will always force things to be different) but most importantly coherence between authors. The tool can make sure that code produced by a team member will look similar to a code produced by another team member, regardless of how they prefer to wrap their lines, indent things, and format code in general.

relrod:
Very begrudgingly merged because I want to get on with my day.

PR #71, opened 2020-01-14
Unit tests

ryanlerch:
a WIP PR, requesting comments.

I tried to stub out the files for the tests, as well as adding a few basic non-authed tests.

tried to split the commits out a bit, but happy to squash (or break them up further if needed)


abompard (changes requested):
The `securitas/tests/utility/` directory should be under `securitas/tests/unit/` like the rest.

abompard:
I've added a commit which integrates VCR and contains one unit tests that demonstrates it.

ryanlerch:
@abompard added some more tests too.

abompard (approved):
Yep I think it's a good structure, we can merge than and improve upon it with more tests.

PR #73, opened 2020-01-15
a simple vagrant setup

ryanlerch:
A dev environment for securitas using vagrant and vagrant libvirt.

Quick Instructions:

**Install Deps & start libvirt**

    sudo dnf install ansible libvirt vagrant-libvirt vagrant-sshfs vagrant-hostmanager
    sudo systemctl enable libvirtd
    sudo systemctl start libvirtd

**Create the environment**

    cd <your-securitas-checkout>
    vagrant up

**Start the server**

    cd <your-securitas-checkout>
    vagrant ssh
    cd /vagrant/
    poetry run flask run --host=0.0.0.0

now in the browser on your host, go to http://ipa.example.com:5000

note too you can go to ipa.example.com as well for the normal IPA webui

### DIFF
--- a/devel/ansible/roles/dev/files/.bashrc
+++ b/devel/ansible/roles/dev/files/.bashrc
@@ -1,0 +1,6 @@
+# .bashrc
+
+export FLASK_APP=/vagrant/securitas/app.py
+export SECURITAS_CONFIG_PATH=/home/vagrant/securitas.cfg
+
+alias securitas-devel="poetry run flask run --host=0.0.0.0"
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -25,6 +25,35 @@ below.
 .. note:: If you do not wish to use GitHub, please send patches to
           infrastructure@lists.fedoraproject.org.
 
+Development Environment
+=======================
+Vagrant allows contributors to get quickly up and running with a Securitas development environment by
+automatically configuring a virtual machine. To get started, first install the Vagrant and Virtualization 
+packages needed, and start the libvirt service::
+
+    $ sudo dnf install ansible libvirt vagrant-libvirt vagrant-sshfs vagrant-hostmanager
+    $ sudo systemctl enable libvirtd
+    $ sudo systemctl start libvirtd
+
+Check out the code and run ``vagrant up``::
+
+    $ git clone https://github.com/fedora-infra/securitas
+    $ cd securitas
+    $ vagrant up
+
+Next, SSH into your newly provisioned development environment, and start the securitas web application:
+
+    $ vagrant ssh
+    $ cd /vagrant/
+    $ poetry run flask run --host=0.0.0.0
+
+With the securitas web application running, go to http://ipa.example.com:5000/ in the browser on your
+host machine to see the web application. http://ipa.example.com will give you access to the regular freeIPA
+webUI.
+
+Note that the ``/vagrant/`` file contains the source of the git checkout on your host. Any changes
+to the files in that directory on the host will be automatically synced to the VM.
+
 
 Guidelines
 ==========
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ docs/_build/
 
 # VS Code
 .vscode
+
+# Vagrant
+.vagrant
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,25 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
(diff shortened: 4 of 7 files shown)


ryanlerch:
Updated with some tweaks to it might work now

abompard:
Works for me! Could you add `.vagrant` to the `.gitignore` please?

devel/ansible/roles/dev/files/.bashrc
```diff
@@ -0,0 +1,6 @@
+# .bashrc
+
+export FLASK_APP=/vagrant/securitas/app.py
+export SECURITAS_CONFIG_PATH=/home/vagrant/securitas.cfg
+
+alias securitas-devel="poetry run flask run --host=0.0.0.0 -h0"
```
relrod: the two flags here do the same thing, only one of them is needed.

devel/ansible/roles/dev/files/securitas.cfg
```diff
@@ -0,0 +1,22 @@
+TEMPLATES_AUTO_RELOAD = True
```
relrod: Is this whole file worth duplicating? Can it use the one from the root of the repo instead?

relrod (dismissed):

StephenCoady:
PR created here for the script for dummy data: https://github.com/ryanlerch/securitas/pull/1

Also adde the vagrant folder to `.gitignore`

devel/ansible/roles/dev/files/.bashrc
ryanlerch: oh, *facepalm* thanks @relrod

devel/ansible/roles/dev/files/securitas.cfg
ryanlerch: okies, tweaked this, removing the duplicated file, and changed the default cfg a bit to suit, since these appeared to be dummy values anyways.

also, changed the password we add to the ipa server in vagrant to match what was already in the default cfg.

ryanlerch:
okies, i think that is most changes fixed now. 

just need to update the developer docs.

@StephenCoady we might get this one in, then work on merging your dummy data one later?

ryanlerch:
Okay, added some documentation.

@relrod @abompard @StephenCoady this is ready for re-review now

docs/contributing.rst
```diff
@@ -25,6 +25,35 @@ below.
...
+
+    $ git clone https://github.com/fedora-infra/securitas
+    $ cd securitas
+    $ vagrant up
+
+Next, SSH into your newly provisioned development environment, and start the securitas web application:
+
+    $ vagrant ssh
+    $ cd /vagrant/
+    $ poetry run flask run --host=0.0.0.0
```
abompard, line 48: You could instruct devs to just use `securitas-devel` here, no?

docs/contributing.rst
```diff
@@ -25,6 +25,35 @@ below.
...
+
+With the securitas web application running, go to http://ipa.example.com:5000/ in the browser on your
+host machine to see the web application. http://ipa.example.com will give you access to the regular freeIPA
+webUI.
+
+Note that the ``/vagrant/`` file contains the source of the git checkout on your host. Any changes
```
abompard, line 54: The `/vagrant/` directory?

abompard (dismissed):
Also, please add `.vagrant` to the `.gitignore`.

docs/contributing.rst
StephenCoady: I would go one further and say that we should probably just start securitas as part of the vagrant box coming up. If we want to go with this, for now, I can add that to a separate PR.

abompard: We can, but in any case we should tell devs how to run the server, in case they want to stop it to fix something and restart it later on.

abompard:
One more thing: you could add `cd /vagrant` at the end of the `.bash_profile` so we get to the code directory immediately after ssh'ing into the box.

devel/ansible/roles/dev/files/.bashrc
```diff
@@ -0,0 +1,6 @@
+# .bashrc
+
+export FLASK_APP=/vagrant/securitas/app.py
+export SECURITAS_CONFIG_PATH=/home/vagrant/securitas.cfg
+
+alias securitas-devel="poetry run flask run --host=0.0.0.0"
```
abompard, line 6: Maybe add one more alias to run the tests?
```
alias securitas-unit-tests="SECURITAS_CONFIG_PATH=/vagrant/securitas/tests/unit/securitas.cfg poetry run pytest -vv securitas/tests/unit"
```

StephenCoady (approved):

PR #74, opened 2020-01-15
fix incorrect FREEIPA_SERVER config key

ryanlerch:
previously, ipa_admin was trying to reference a
non-existant configuration key "FREEIPA_SERVER".
This commit changes the key to the correct "FREEIPA_SERVERS",
and uses the same random technique used in security/ipa.py
to choose a single server from the list in FREEIPA_SERVERS

### DIFF
--- a/securitas/security/ipa_admin.py
+++ b/securitas/security/ipa_admin.py
@@ -1,4 +1,5 @@
 from python_freeipa import Client
+import random
 
 
 class IPAAdmin(object):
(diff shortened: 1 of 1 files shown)


abompard (approved):
Looks good to me but I'd like somebody else's approval too since it's a file in the `security` directory. @relrod could you check?

relrod:
LGTM and thank you for the two-phase review on security files.

PR #79, opened 2020-01-17
Registration form: handle some error cases

abompard:
Handle some error cases that can happen during registration:
- The password does not fit the policy
- The username is already taken
- The username contains invalid characters

### DIFF
--- a/securitas/controller/registration.py
+++ b/securitas/controller/registration.py
@@ -27,6 +27,23 @@ def register():
             # This also logs the user in right away.
             ipa = untouched_ipa_client(app)
             ipa.change_password(username, password, password)
+        except python_freeipa.exceptions.PWChangePolicyError as e:
+            # situations like "password is too short"
+            form.password.errors.append(e.policy_error)
+            ipa_admin.user_del(username)
+        except python_freeipa.exceptions.DuplicateEntry as e:
+            # the username already exists
+            form.username.errors.append(e.message)
+        except python_freeipa.exceptions.ValidationError as e:
+            # for example: invalid username. We don't know which field to link it to
+            if e.message.startswith("invalid 'login': "):
+                form.username.errors.append(e.message[len("invalid 'login': ") :])
+            else:
+                app.logger.error(
+                    f'An unhandled invalid value happened while registering user '
+                    f'{username}: {e.message}'
+                )
+                form.errors['non_field_errors'] = [e.message]
         except python_freeipa.exceptions.FreeIPAError as e:
             app.logger.error(
                 f'An unhandled error {e.__class__.__name__} happened while registering user '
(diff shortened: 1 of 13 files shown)


StephenCoady (approved):
@abompard I've validated all three cases in your description. Works a charm! Happy to merge this as is

securitas/controller/registration.py
```diff
@@ -1,48 +1,63 @@
...
+            )
+
+            # Now we fake a password change, so that it's not immediately expired.
+            # This also logs the user in right away.
+            ipa = untouched_ipa_client(app)
+            ipa.change_password(username, password, password)
+        except python_freeipa.exceptions.PWChangePolicyError as e:
+            # situations like "password is too short"
+            form.password.errors.append(e.policy_error)
+            ipa_admin.user_del(username)
```
relrod, line 33: @abompard

Sorry, but what problem is this (and the one below) solving? What happens if I give a username that already exists, but also a short password, are you 150% sure that `DuplicateEntry` gets thrown before `PWChangePolicyError` in *every* possible case or is there a risk that the user account gets deleted? Are you 150% sure (below) that I can't trigger a `FreeIPAError` before a `DuplicatEntry` based one some bad field entry I make?

I think we should *never* `user_del` *anywhere* in the codebase, and I think it should be VERY rare that we use `ipa_admin` (which, in production, will not have permission to `user_del` anyway, it will be locked down to have only the ability to change passwords and create accounts).

I believe we should remove the `user_del`.

abompard: > Sorry, but what problem is this (and the one below) solving?

Before this change, when a user registers with a password that does not follow the policy, they get an error message about being unable to register. But the user is actually created, it's just the password change that failed, so it's pretty confusing. 

As for the one below (I suppose you mean catching the generic `FreeIPAError`), it's about catching other errors that happen during the password change process and not leaving the user in a similar state.

> What happens if I give a username that already exists, but also a short password, are you 150% sure that `DuplicateEntry` gets thrown before `PWChangePolicyError` in _every_ possible case or is there a risk that the user account gets deleted?

Fair question, the `DuplicateEntry` exception will be raised by the `user_add` method, and only `change_password` raises a `PWChangePolicyError`, so you can't reach that error handler on an existing account.

> Are you 150% sure (below) that I can't trigger a `FreeIPAError` before a `DuplicatEntry` based one some bad field entry I make?

Ah, I did not think of that one, and `FreeIPAError` is indeed a pretty generic error so I think it could happen (and if not now it may happen in future evolutions of the library). I agree that this one should go.
It's probably a bad idea to have common error handlers for both method calls anyway.

> I think we should _never_ `user_del` _anywhere_ in the codebase, and I think it should be VERY rare that we use `ipa_admin` (which, in production, will not have permission to `user_del` anyway, it will be locked down to have only the ability to change passwords and create accounts).

That's an interesting piece of information indeed :-) Then we must find a way to not leave users out in the cold when they pick a password that does not comply with the policy. I'll think about it and send a PR.

Thanks for the comments.

abompard: OK I've sent #93, please tell me what you think.

PR #80, opened 2020-01-17
Vagrant

StephenCoady:
added a securitas service which autostarts.
also added a python script which auto generates some user and group data


devel/ansible/roles/dev/files/.bashrc
```diff
@@ -3,4 +3,10 @@
 export FLASK_APP=/vagrant/securitas/app.py
 export SECURITAS_CONFIG_PATH=/home/vagrant/securitas.cfg
 
-alias securitas-devel="poetry run flask run --host=0.0.0.0"
\ No newline at end of file
+alias securitas-start="sudo systemctl start securitas.service 'The Application is running on http://0.0.0.0:5000'"
```
abompard: What does the last text argument do?

devel/ansible/roles/dev/tasks/main.yml
```diff
@@ -40,5 +40,27 @@
...
+      dest: /etc/systemd/system/{{ item }}
+      mode: 0644
+  with_items:
+      - securitas.service
+
+- name: Start and enable the securitas services
+  service:
+      name: "{{ item }}"
+      state: started
+      enabled: yes
```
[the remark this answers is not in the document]
StephenCoady: Done.

devel/ansible/roles/dev/files/.bashrc
StephenCoady: I forgot to add echo to each, just providing feedback as systemctl doesn't.

devel/ansible/roles/dev/tasks/main.yml
```diff
@@ -57,10 +57,9 @@
...
 
-- name: Start and enable the securitas services
-  service:
-      name: "{{ item }}"
-      state: started
-      enabled: yes
-  with_items:
-      - securitas
\ No newline at end of file
+- name: Reload systemd so it learns about securitas
```
abompard: The name does not describe what the task does, the task also starts the daemon (and that's the main thing, I would say)

abompard (changes requested):

docs/contributing.rst
```diff
@@ -41,17 +41,23 @@ Check out the code and run ``vagrant up``::
...
 
     $ vagrant ssh
-    $ cd /vagrant/
-    $ poetry run flask run --host=0.0.0.0
 
-With the securitas web application running, go to http://ipa.example.com:5000/ in the browser on your
+where you can run the following commands:
+
+    $ securitas-restart
+		$ securitas-stop
```
abompard: Indentation seems weird here, maybe tabs instead of spaces ?

tools/create-test-data.py
```diff
@@ -0,0 +1,49 @@
...
+try:
+    ipa.group_add("developers", "A group for developers")
+except python_freeipa.exceptions.FreeIPAError as e:
+    print(e)
+
+# create some random users and add them to both the developers and admin groups randomly
+for x in range(50):
+    firstName = fake.first_name()
+    lastName = fake.last_name()
+    username = firstName + str(x)
```
abompard: can `first_name()` generate a string with spaces? The `username` value has more constraints than the first name.

tools/create-test-data.py
```diff
@@ -0,0 +1,49 @@
...
+    try:
+        ipa.user_add(
+            username,
+            firstName,
+            lastName,
+            firstName + " " + lastName,
+            home_directory='/home/fedora/%s' % username,
```
abompard: You can also use `f"/home/fedora/{username}"` to format the string (since Python 3.6).

tools/create-test-data.py
```diff
@@ -0,0 +1,49 @@
...
+            disabled=False,
+            random_pass=True,
+            fasircnick=username,
+            faslocale=None,
+            fastimezone=None,
+            fasgpgkeyid=[None],
```
abompard: This should be a list with `None` as only element? Seems weird, why not an empty list?

tools/create-test-data.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+        users.append(username)
+        for i in range(0,len(users)):
```
abompard: You can also use:
```
for i, user in enumerate(users):
```

tools/create-test-data.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
+
+import python_freeipa
+from python_freeipa import Client
+import random
```
abompard: Traditionnaly, imports from the std library come first, and are separated from other imports by a blank line.

tools/create-test-data.py
```diff
@@ -0,0 +1,49 @@
...
+from faker import Faker
+fake = Faker()
```
abompard: I'm surprised Black didn't insert two lines between the last imports and the first code line... Did you run Black on this file?

tools/create-test-data.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
```
abompard: There's currently a `devel` directory with the ansible conf for Vagrant, maybe this script could be put in that directory instead of creating `tools`?

abompard (changes requested):

tools/create-test-data.py
StephenCoady: I actually wasn't sure of these data structures in general. I left them as `None` for now but they can be changed.

tools/create-test-data.py
StephenCoady: I don't think so, but if it does it should fail gracefully anyway

tools/create-test-data.py
abompard: Yeah I think it's better to have an empty list than a list with a single `None` element.

devel/ansible/roles/dev/files/create-test-data.py
```diff
@@ -0,0 +1,51 @@
...
+            random_pass=True,
+            fasircnick=username,
+            faslocale=None,
+            fastimezone=None,
+            fasgpgkeyid=[None],
+        )
+        users.append(username)
+        for i, user in enumerate(users):
+            if i % 3 == 0 and i!= 0:
+                ipa.group_add_member("developers", users[i], skip_errors=True)
```
abompard: now that you have `user` in your loop (thanks to `enumerate()`), you can just use it and you don't need `users[i]`.

devel/ansible/roles/dev/files/create-test-data.py
```diff
@@ -0,0 +1,51 @@
...
+            if i % 2 == 0 and i!= 0:
+                ipa.group_add_member("admins", users[i], skip_errors=True)
```
abompard: same here

devel/ansible/roles/dev/files/create-test-data.py
```diff
@@ -0,0 +1,51 @@
...
+            home_directory=f"/home/fedora/{username}",
+            disabled=False,
+            random_pass=True,
+            fasircnick=username,
+            faslocale=None,
+            fastimezone=None,
+            fasgpgkeyid=[None],
+        )
+        users.append(username)
+        for i, user in enumerate(users):
```
abompard: I don't understand why you need to loop over the `users` list each time you create a user. Why not use `x` from the top loop?

abompard (changes requested):

tools/create-test-data.py
relrod: +1

devel/ansible/roles/dev/files/create-test-data.py
StephenCoady: fair point, changed now

devel/ansible/roles/dev/files/securitas.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=securitas
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Environment=FLASK_APP=/vagrant/securitas/app.py
+Environment=SECURITAS_CONFIG_PATH=/home/vagrant/securitas.cfg
```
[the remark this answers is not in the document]
StephenCoady: good idea, done. thanks

devel/ansible/roles/dev/tasks/main.yml
```diff
@@ -40,5 +40,26 @@
...
       group: vagrant
 
+- name: Install dummy data deps
+  pip:
+    name:
+      - python-freeipa
+      - Faker
 
+- name: Create dummy freeipa data
+  shell: python3 /vagrant/devel/ansible/roles/dev/create-test-data.py
```
abompard: I think this could have been directly in `/devel` so it's more easily accessible in vagrant itself, and because it has nothing to do with ansible (it's not copied through ansible).

abompard (commented):
Just a minor nitpick, then it's good to go :-)

abompard (approved):
Looks good, thanks!

PR #81, opened 2020-01-17
Add tests to the current securitas codebase

abompard:
Use WTForm for the login form and add tests on the login and logout controllers.


securitas/controller/authentication.py
```diff
@@ -1,43 +1,55 @@
...
+from securitas.form.login_user import LoginUserForm
+from securitas.security.ipa import maybe_ipa_login, maybe_ipa_session
 
 
 @app.route('/logout')
-@with_ipa(app, session)
-def logout(ipa):
+def logout():
+    """Log the user out."""
+    # Don't use the with_ipa() decorator, otherwise anonymous users visiting this endpoint will be
```
relrod, line 12: Heh, fair. I guess I didn't consider that an issue because an anonymous user shouldn't be going to this endpoint. :laughing:

securitas/templates/login.html
```diff
@@ -0,0 +1,26 @@
+{% extends "with_navbar.html" %}
+{% block title %}Self-Service Portal{% endblock %}
```
relrod: Perhaps use a more descriptive title that mentions "login" somewhere?

securitas/form/login_user.py
```diff
@@ -0,0 +1,13 @@
+from flask_wtf import FlaskForm
+from wtforms import StringField, PasswordField
+from wtforms.validators import DataRequired
+
+
+class LoginUserForm(FlaskForm):
```
relrod, line 6: Is it worth trying to encode some rules here about what constitutes a valid username/password? We'll basically be re-encoding the IPA rules I guess, but the advantage is that some of them could be verified on the client side before the form is ever submitted (e.g. min/max length).

securitas/controller/authentication.py
```diff
@@ -1,43 +1,55 @@
...
+            ipa = maybe_ipa_login(app, session, username, password)
+        except python_freeipa.exceptions.PasswordExpired:
+            flash('Password expired. Please reset it.', 'red')
+            return redirect(url_for('password_reset'))
+        except python_freeipa.exceptions.Unauthorized as e:
+            form.errors['non_field_errors'] = [e.message]
+        except python_freeipa.exceptions.FreeIPAError as e:
+            # If we made it here, we hit something weird not caught above. We didn't
+            # bomb out, but we don't have IPA creds, either. Boot us back to /.
+            app.logger.error(
```
relrod, line 38: I *think* this is okay because the IPA server should never include passwords in the response, but we should be mindful to make sure nothing we log will ever include a password.

relrod (commented):
Some inlines but overall this looks fine to me.

securitas/form/login_user.py
abompard: Yeah I asked myself that question (for the registration form), and I thought that if we want to change the password policy we'd have to do it in two different places, which is error-prone.
So I've only left basic "not-empty" validators.

securitas/controller/authentication.py
abompard: Oh good point, I didn't think of that. Hopefully they don't put passwords in the `message` attribute of their exceptions, though.

securitas/templates/login.html
abompard: Will do.

securitas/tests/unit/conftest.py
```diff
@@ -52,18 +52,49 @@ def vcr_cassette_dir(request):
...
+            user_password=password,
             login_shell='/bin/bash',
+            fascreationtime=f"{now.isoformat()}Z",
         )
         ipa = untouched_ipa_client(app)
-        ipa.change_password('dummy', 'dummy_password', 'dummy_password')
-        yield
-    finally:
-        ipa_admin.user_del('dummy')
+        ipa.change_password(name, password, password)
```
StephenCoady, line 71: Question - why is this line necessary? Is it a temporary password if created using the API?

securitas/tests/unit/controller/test_authentication.py
```diff
@@ -1,71 +1,171 @@
...
+# Login
 
 
 @pytest.mark.vcr()
 def test_login(client):
     """Test a successful Login"""
     result = client.post(
         '/login',
-        data={"username": "dudemcpants", "password": "n:nPv{P].9}]!q$RE%w<38@",},
+        data={"username": "dudemcpants", "password": "n:nPv{P].9}]!q$RE%w<38@"},
```
StephenCoady: shouldn't this be the dummy user now? the cassettes might be hiding this failing.

abompard: Correct, it's only working because we didn't update the cassettes. Also, it should use the `dummy_user` fixture.

securitas/tests/unit/conftest.py
abompard: When a user is create with the API, it's created with an expired password, so they are forced to change it before they can login. We don't want that, so we're changing the password ourselves. It's also what the registration controller does.

StephenCoady (approved):

PR #83, opened 2020-01-27
Restore the Flask dev server's ability to run non-installed apps

abompard:
The flask dev server doesn't need the app to be installed to run it. Changeset 321d0f9 broke that feature, this commit should restore it.


relrod:
Thank you!

PR #87, opened 2020-01-28
Use a newer version of poetry to fix CI

abompard:
Please don't merge just now, I'd like to see if Centos CI picks it up.


abompard:
Actually, let's merge that now

PR #89, opened 2020-01-28
Fix errors in the test data creation script

abompard:
Not much more to say... ;-)


abompard:
I'm going to merge this because the script does not currently work, so it has to be an improvement ;-)

PR #90, opened 2020-01-28
Add unit tests for the password controller

StephenCoady:


securitas/tests/unit/conftest.py
```diff
@@ -83,6 +83,12 @@ def dummy_user(make_user):
     yield
 
 
+@pytest.fixture
+def cleanup_dummy_user():
+    yield
+    ipa_admin.user_del('dummy')
```
abompard: Same comment as on the other PR: I'm not sure it's a good idea to move fixtures to the shared module if they are not used by more than one test module. I'm afraid it may end up cluttering the `conftest.py` file.

securitas/tests/unit/controller/test_password.py
```diff
@@ -9,3 +11,117 @@ def test_password_reset(client):
...
+            "password": "LongSuperSafePassword",
+            "password_confirm": "LongSuperSafePassword",
+        },
+        follow_redirects=True,
+    )
+    assert result.status_code == 200
+    page = BeautifulSoup(result.data, 'html.parser')
+    messages = page.select(".flash-messages .red")
+    assert len(messages) == 1
+    assert messages[0].get_text(strip=True) == 'Failed to reset your password (invalid current password).'
```
abompard: Is this a password policy thing? This test is called `test_password_policy` but it looks like it's only testing that the current password is correct, not that the new passwords follows the policy. It looks identical to `test_incorrect_current_password`.

securitas/tests/unit/controller/test_password.py
```diff
@@ -9,3 +11,117 @@ def test_password_reset(client):
...
+            "password": "1",
+            "password_confirm": "1",
+        },
+        follow_redirects=True,
+    )
+    assert result.status_code == 200
+    page = BeautifulSoup(result.data, 'html.parser')
+    messages = page.select(".flash-messages .red")
+    assert len(messages) == 1
+    assert messages[0].get_text(strip=True) == 'Failed to reset your password (policy error): Constraint violation: Too soon to change password'
```
abompard: This does not look like the error message we want to get in this situation.

securitas/tests/unit/controller/test_password.py
```diff
@@ -9,3 +11,117 @@ def test_password_reset(client):
...
+        follow_redirects=True,
+    )
+    assert result.status_code == 200
+    page = BeautifulSoup(result.data, 'html.parser')
+    messages = page.select(".flash-messages .red")
+    assert len(messages) == 1
+    assert messages[0].get_text(strip=True) == 'Password and confirmation did not match.'
+
+
+def test_password_policy(client):
```
abompard: You should probably use the `dummy_user` fixture, so you don't depend on a specific user existing in FreeIPA before the tests are run.
Also, this tests needs the VCR decorator.

securitas/tests/unit/controller/test_password.py
```diff
@@ -9,3 +11,117 @@ def test_password_reset(client):
...
+        follow_redirects=True,
+    )
+    assert result.status_code == 200
+    page = BeautifulSoup(result.data, 'html.parser')
+    messages = page.select(".flash-messages .red")
+    assert len(messages) == 1
+    assert messages[0].get_text(strip=True) == 'Failed to reset your password (policy error): Constraint violation: Too soon to change password'
+
+@pytest.mark.vcr()
+def test_incorrect_current_password(client):
```
abompard: You should probably use the `dummy_user` fixture, so you don't depend on a specific user existing in FreeIPA before the tests are run.

securitas/tests/unit/controller/test_password.py
```diff
@@ -9,3 +11,117 @@ def test_password_reset(client):
...
+        follow_redirects=True,
+    )
+    assert result.status_code == 200
+    page = BeautifulSoup(result.data, 'html.parser')
+    messages = page.select(".flash-messages .red")
+    assert len(messages) == 1
+    assert messages[0].get_text(strip=True) == 'Failed to reset your password (invalid current password).'
+
+@pytest.mark.vcr()
+def test_password_changes(client):
```
abompard: You should probably use the `dummy_user` fixture, so you don't depend on a specific user existing in FreeIPA before the tests are run.

abompard (dismissed):

abompard:
Ugh, I was commenting on a single commit, I'll remove my review and start over.

securitas/templates/password-reset.html
```diff
@@ -22,24 +24,20 @@
...
 
               <div class="input-field col s12">
-                <input placeholder="password" id="password_confirm" name="password_confirm" type="password" class="validate" tabindex="4">
-                <label for="password_confirm">Password Confirmation</label>
-              </div>
+                {{ macros.with_errors(password_reset_form.password_confirm, class="validate", tabindex="4")}}
+                </div>
 
               <div class="input-field col s12">
                 <button class="btn waves-effect waves-light light-blue darken-4 w100" id="submit" type="submit" tabindex="5">Submit</button>
```
abompard, line 43: You'll also need to add the macro that displays the `non_field_errors` errors, so something like:
```
{{ macros.non_field_errors(password_reset_form) }}
```

securitas/tests/unit/conftest.py
```diff
@@ -83,6 +83,26 @@ def dummy_user(make_user):
...
+def dummy_user_as_group_manager(logged_in_dummy_user, dummy_group):
+    """Make the dummy user a manager of the dummy-group group."""
+    ipa_admin.group_add_member("dummy-group", users="dummy")
+    ipa_admin.group_add_member_manager("dummy-group", users="dummy")
+    yield
+
+
+@pytest.fixture
+def remove_password_min_time():
+    ipa_admin.pwpolicy_mod("dummy-group", 0)
```
abompard: I think it'd be clearer and more future-proof if you used the keyword here for the second argument (`krbminpwdlife=0`).
Also, I think it would be better to restore the value on teardown (after a `yield`) so that it does not affect other tests running after the test that has used this fixture. It can look kinda useless since now we don't see why we would ever need the krbminpwlife to be more than zero, but I think it's a good practice for the testsuite to leave the directory in the state that we found it in.

securitas/tests/unit/controller/test_password.py
```diff
@@ -9,3 +10,86 @@ def test_password_reset(client):
...
+            "password": "1",
+            "password_confirm": "1",
+        },
+        follow_redirects=True,
+    )
+    page = BeautifulSoup(result.data, 'html.parser')
+    password_input = page.select("input[name='password']")[0]
+    assert 'invalid' in password_input['class']
+    helper_text = password_input.find_next("span", class_="helper-text")
+    assert helper_text["data-error"] == "Constraint violation: Too soon to change password"
```
abompard: This does not look like the error we are expecting in this situation. I would expect something like "password is too short".

securitas/tests/unit/controller/test_password.py
```diff
@@ -9,3 +10,86 @@ def test_password_reset(client):
...
+
+
+@pytest.mark.vcr()
+def test_password_changes(client, dummy_user,
+                          dummy_user_as_group_manager, remove_password_min_time):
```
abompard: The `dummy_user_as_a_group_manager` fixture already depends on the `dummy_user` fixture, so you don't need to list it here.

securitas/tests/unit/conftest.py
```diff
@@ -83,6 +83,26 @@ def dummy_user(make_user):
...
+@pytest.fixture
+def dummy_user_as_group_manager(logged_in_dummy_user, dummy_group):
+    """Make the dummy user a manager of the dummy-group group."""
+    ipa_admin.group_add_member("dummy-group", users="dummy")
+    ipa_admin.group_add_member_manager("dummy-group", users="dummy")
+    yield
+
+
+@pytest.fixture
+def remove_password_min_time():
```
abompard: Until now the fixtures have been named after the state they create, not as "actions", what would you say about renaming this one to something like "no_password_min_time"? (very very minor nitpick)

abompard (changes requested):

securitas/tests/unit/conftest.py
StephenCoady: Yeah, the reason I didn't reset it was because the group gets deleted by dummy_group anyway. I will add it though just for completeness.

securitas/tests/unit/controller/test_password.py
StephenCoady: You are somewhat correct. It looks like the time constraint takes preference over a short password and as soon as that is triggered it doesn't matter what they tried to change it to. I will add another test with the time constraint removed and a short password.

securitas/tests/unit/conftest.py
abompard: Oh, I thought it removed the limit for everyone, but indeed it's only for the dummy-group... Then you should depend on the `dummy_group` fixture, and yeah it's not as necessary to remove it afterwards. Up to you :-)

abompard: OK now I get it, you actually want to add a password policy for the dummy-group where the minlifetime is zero. The command you want to use then is `pwpolicy_add`, not `pwpolicy_mod`.
I suggest the following changes:
```patch
--- a/securitas/security/ipa.py
+++ b/securitas/security/ipa.py
@@ -46,7 +46,7 @@ class Client(IPAClient):
             parse_group_management_error(data)
         return data['result']
 
-    def pwpolicy_mod(
+    def pwpolicy_add(
         self, group, krbminpwdlife=None, **kwargs
     ):
         """
@@ -57,7 +57,7 @@ class Client(IPAClient):
         """
         params = {'all': True, 'raw': True, 'krbminpwdlife': krbminpwdlife}
         params.update(kwargs)
-        data = self._request('pwpolicy_mod', group, params)
+        data = self._request('pwpolicy_add', group, params)
         return data['result']
 
 
--- a/securitas/security/ipa_admin.py
+++ b/securitas/security/ipa_admin.py
@@ -13,7 +13,7 @@ class IPAAdmin(object):
         "group_del",
         "group_add_member",
         "group_add_member_manager",
-        "pwpolicy_mod"
+        "pwpolicy_add",
     )
 
     def __init__(self, app):
--- a/securitas/tests/unit/conftest.py
+++ b/securitas/tests/unit/conftest.py
@@ -100,7 +100,7 @@ def dummy_user_as_group_manager(logged_in_dummy_user, dummy_group):
 
 @pytest.fixture
 def remove_password_min_time():
-    ipa_admin.pwpolicy_mod("dummy-group", 0)
+    ipa_admin.pwpolicy_add("dummy-group", krbminpwdlife=0, cospriority=10)
 
 
 @pytest.fixture
```

pyproject.toml
```diff
@@ -35,6 +35,8 @@ liccheck = "^0.3.12"
 bandit = "^1.6"
 black = {version = "^18.3-alpha.0", allow-prereleases = true}
 flake8 = "^3.7"
+python-freeipa = "^1.0"
```
abompard: This is not a development dependency, it's already present in the main dependencies.

pyproject.toml
```diff
@@ -35,6 +35,8 @@ liccheck = "^0.3.12"
...
+Faker = "^4.0.0"
```
abompard, line 38: Whenever you change the `pyproject.toml` file, you have to run `poetry lock` afterwards so it updates the lock file. Or you can use `poetry add -D Faker` instead of editing the file, it will do both.

securitas/security/ipa.py
```diff
@@ -46,6 +46,21 @@ def group_add_member_manager(
             parse_group_management_error(data)
         return data['result']
 
+    def pwpolicy_add(
+        self, group, krbminpwdlife=None, krbpwdminlength=None, cospriority=None, **kwargs
+    ):
+        """
+        Create the password time policy
```
abompard: Actually this method can do more than create a password *time* policy, it can create the whole password policy given the proper kwargs (even if they are not explicitely listed yet). I would suggest removing `time`.

securitas/tests/unit/controller/test_password.py
```diff
@@ -9,3 +10,105 @@ def test_password_reset(client):
...
+    assert result.status_code == 200
+    page = BeautifulSoup(result.data, 'html.parser')
+    password_input = page.select("input[name='password']")[0]
+    assert 'invalid' in password_input['class']
+    helper_text = password_input.find_next("span", class_="helper-text")
+    assert helper_text["data-error"] == "Passwords must match"
+
+
+@pytest.mark.vcr()
+def test_password(client):
```
abompard: This should probably use the `dummy_user` fixture to avoid depending on something that existed in FreeIPA before the test run.

securitas/tests/unit/controller/test_password.py
```diff
@@ -9,3 +10,105 @@ def test_password_reset(client):
...
+            "password": "somesupersecretpassword",
+            "password_confirm": "somesupersecretpassword",
+        },
+        follow_redirects=True,
+    )
+    page = BeautifulSoup(result.data, 'html.parser')
+    password_input = page.select("input[name='password']")[0]
+    assert 'invalid' in password_input['class']
+    helper_text = password_input.find_next("span", class_="helper-text")
+    assert helper_text["data-error"] == "Constraint violation: Too soon to change password"
```
abompard: That's going to depend on the order in which the tests are run: some other test must have changed the password before. Not sure how to work around that, though.

securitas/tests/unit/conftest.py
```diff
@@ -83,6 +83,26 @@ def dummy_user(make_user):
...
+@pytest.fixture
+def dummy_user_as_group_manager(logged_in_dummy_user, dummy_group):
+    """Make the dummy user a manager of the dummy-group group."""
+    ipa_admin.group_add_member("dummy-group", users="dummy")
+    ipa_admin.group_add_member_manager("dummy-group", users="dummy")
+    yield
+
+
+@pytest.fixture
+def no_password_min_time(dummy_user_as_group_manager):
```
abompard: You don't actually need the `dummy_user_as_group_manager fixture`, no? It looks like you only need the `dummy_group` fixture.

securitas/tests/unit/controller/test_password.py
```diff
@@ -9,3 +10,105 @@ def test_password_reset(client):
...
+
+
+@pytest.mark.vcr()
+def test_short_password(client, no_password_min_time):
```
abompard: If you change the dependency of `no_password_min_time`, you'll have to add a dependency on `dummy_user` here.

securitas/tests/unit/controller/test_password.py
```diff
@@ -9,3 +10,105 @@ def test_password_reset(client):
...
+    )
+    page = BeautifulSoup(result.data, 'html.parser')
+    password_input = page.select("input[name='password']")[0]
+    assert 'invalid' in password_input['class']
+    helper_text = password_input.find_next("span", class_="helper-text")
+    assert helper_text["data-error"] == "Constraint violation: Password is too short"
+
+
+@pytest.mark.vcr()
+def test_password_changes(client, no_password_min_time):
```
abompard: Same thing here, you'll need `dummy_user` if you change the deps of the other fixture.

securitas/tests/unit/controller/test_password.py
```diff
@@ -9,3 +10,105 @@ def test_password_reset(client):
...
+            "password": "secretpw",
+            "password_confirm": "secretpw",
+        },
+        follow_redirects=True,
+    )
+    assert result.status_code == 200
+    page = BeautifulSoup(result.data, 'html.parser')
+    messages = page.select(".flash-messages .green")
+    assert len(messages) == 1
+    assert messages[0].get_text(strip=True) == \
```
abompard: In Python it's more common to use the parenthesis than the backslash to handle line breaks, so you'd have something like:
```python
assert messages[0].get_text(strip=True) == (
    'Your password has been changed, please try to log in with the new one now.'
)
```
But then again, if Black is happy with it I have nothing to say ;-)

abompard (changes requested):
Getting there! A couple comments again. Coverage is also reporting a couple line as not covered by the tests (the `FreeIPAError` handler). Thanks a lot for your work!

securitas/tests/unit/controller/test_password.py
StephenCoady: I think this works because in the fixtures file you have `ipa.change_password()` right after the dummy user is created. It works well because its also the reason I needed to disable the time policy further down. The only other thing I can think of is disable the time policy, change the password again, re-enable the time policy and then run this test.

securitas/tests/unit/controller/test_password.py
StephenCoady: Ran black again, made some changes. Thanks!

securitas/tests/unit/controller/test_password.py
abompard: Ah good point, the password is always changed after user creation. Could you add a comment to explain that please?

securitas/tests/unit/controller/test_password.py
```diff
@@ -1,4 +1,8 @@
+import pytest
+import python_freeipa
 from bs4 import BeautifulSoup
+from unittest import mock
+from securitas import app
```
abompard: Flake8 says that this variable is unused, please remove it.

abompard (changes requested):
Just one small thing caught by flake8, after that it's good to go! :-)

abompard (approved):
Looks good, thanks a lot!

PR #93, opened 2020-01-29
Don't delete the user when registration fails

abompard:
This will avoid potential attack vectors, as well as actually work in production where we won't be allowed to delete users anyway.

When a password does not comply with policy, it will stay expired and the newly registered user will be sent to the login page where they will have to change it.

I see one downside: during the time between registration and the password change, the user will have an expired password that does not comply with the policy (and might very well be very insecure). Right now I don't know how to mitigate that.

I'd love to have the opinion from an UX person (wink wink @ryanlerch) on the workflow I've implemented here. To test it, try to register with too short a password.

Also pinging @relrod who raised the initial issue.

### DIFF
--- a/securitas/controller/registration.py
+++ b/securitas/controller/registration.py
@@ -18,6 +18,7 @@ def register():
         username = form.username.data
         password = form.password.data
         now = datetime.datetime.utcnow().replace(microsecond=0)
+        # First, create the user.
         try:
             ipa_admin.user_add(
                 username,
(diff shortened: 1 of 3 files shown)


sfinn85:
Hi @ryanlerch 
@relrod just commented on email to ask if you can review this from UX perspective, following @abompard  comment, thanks

ryanlerch:
i'm a little confused here as to why a user actually gets created if registration fails in the first place.

sfinn85:
Ryan and @abompard to chat further to resolve

abompard:
I'll explain here in writing: registering is a two steps process.

1. The user is created in FreeIPA with the given password, but by FreeIPA policy this password is automatically set as "expired", and the user must change it upon login. At that point the passwords policies (length, complexity, etc) don't apply since an admin is setting the password.
2. To work around the password being expired, the backend logs in as the user and does a change password operation to re-set the password to what the user has requested. This time the user is setting their password, so complexity policies apply.

If the user has set a password that does not conform to policy (e.g it's too short), the first operation will succeed and the second operation will fail, leaving us with the account created in FreeIPA but with an expired password.

What I first thought is that we could delete the user we just created and start over, but there is a security risk in deleting users, and we won't be able to do that in production anyway. So I'm trying to explain to the user what's going on in case the change password operation is rejected, and tell them that they'll have to change their password upon login.

ryanlerch (approved):

PR #97, opened 2020-02-03
feat: add button to route to pagure for delete

### LINKED ISSUE #39: (M) As a registered user, I want to disable my account, so that I can act within my GDPR rights.
AC:
We have a SAR script to list information about users in our repository
We have a means to link users to the pagure instance where they can request an account disable

DoD:
All prs implementing the above are merged and have been verified

StephenCoady:

### DIFF
--- a/securitas/templates/user-edit.html
+++ b/securitas/templates/user-edit.html
@@ -6,15 +6,15 @@
 {% block content %}
   {{ super() }}
   <div class="container section">
-    <div class="row">
+    <div class="row form-group">
       <div class="col s12">
         <h3 class="header light-blue-text text-darken-4 fw300">
           Edit user: {{ user.username }}
         </h3>
-
+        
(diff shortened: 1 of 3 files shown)


relrod:
Deferring to @ryanlerch for review since this text is Fedora-specific and there was talk with the theming process of making securitas less Fedora specific.

ryanlerch:
@relrod yeah, will add the FAS theme PR today, with some guidance on how to add optional items like this to the theme.

ryanlerch:
The following PR adds a FAS theme, and adds a button for editing group details to it:
https://github.com/fedora-infra/securitas/pull/98

the editing group details is a similar workflow to removing a user (aka filing a ticket), so we could use the same approach here too (if the approach in m,y pending PR is approved :) )

ryanlerch (approved):
Looking good!

Should we squash these two commits before merging though?

PR #106, opened 2020-02-06
add sar script as per gdpr requirements

### LINKED ISSUE #39: (M) As a registered user, I want to disable my account, so that I can act within my GDPR rights.

StephenCoady:
We need a script to retreive information about a given user from the system.

### DIFF
--- a/deployment/scripts/sar.py
+++ b/deployment/scripts/sar.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 import json
 
 import click
--- a/deployment/scripts/sar.py
+++ b/deployment/scripts/sar.py
@@ -1,5 +1,6 @@
-import click
 import json
+
+import click
 from python_freeipa.client_legacy import ClientLegacy
 
 
--- a/devel/ansible/roles/dev/files/sar.py
+++ b/devel/ansible/roles/dev/files/sar.py
@@ -1,0 +1,95 @@
+import click
+import json
+from python_freeipa.client_legacy import ClientLegacy
+
+
+def print_human_readable_format(sar_data):
+    """
+    Print user data in human readable format.
+
+    Args:
+        sar_data (dict): User data to be printed.
+    """
+    header_start = "==========>"
+    header_stop = "<=========="
+    footer = "======================================================="
+    chapter_start = "---->"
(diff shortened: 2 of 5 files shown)


devel/ansible/roles/dev/files/sar.py
```diff
@@ -0,0 +1,119 @@
+import click
+import requests
```
relrod: Any reason in particular for not using `python-freeipa` like everything else?

StephenCoady: Nothing major. We haven't swapped the client yet to the new version and we spoke about maybe letting our intern James take that task on. I saw no real drawback to using `requests` since it is essentially just two methods. I was thinking it would be good for James to start with this file when he migrates the client to the new version.

devel/ansible/roles/dev/files/sar.py
```diff
@@ -0,0 +1,95 @@
+import click
+import json
```
abompard: In Python there is a tradition to separate imports in the following manner:
1. imports from the stdlib
2. blank line
3. imports from third-party libs
4. blank line
5. imports from the project

In this case it would become:
```
import json

import click
from python_freeipa import ...
```

abompard (commented):
Now that I think of it I think this file should not be in the `devel/ansible/roles/dev/files` folder, because it is not specific to the Vagrant setup. I don't see a perfect place for it yet so maybe create a `deployment` folder an put it there? Later we could put example config files there.

deployment/scripts/sar.py
```diff
@@ -0,0 +1,96 @@
+import json
```
abompard, line 3: Please add `#!/usr/bin/env python` at the top of the file and make it executable.

abompard (changes requested):
One last thing and then I think it should be good :-) Thanks!

StephenCoady:
Thanks @abompard, think this is good to go now!

abompard (approved):
Looks good, thanks!

PR #107, opened 2020-02-07
add the ability in the UI to add new GPG Key Ids

### LINKED ISSUE #96: Impossible to add (multiple) GPG key(s)
It's currently impossible in the user form to add one (or many) GPG keys.

ryanlerch:
Adds the ability to add 4 additional GPG Key IDs on every
save of the edit user dialog. This basic solution was chosen
so we can use append_field() from WTForms fieldlists, rather
than some hand-crafted hacky javascript to add new fields on
the fly.

The drawback with this approach is that a user can only add 4
new GPG Keys at a time. More than 4 keys can be added, the user
just has to submit the form, and then add more.

Resolves: #96

### DIFF
--- a/securitas/templates/_form_macros.html
+++ b/securitas/templates/_form_macros.html
@@ -1,24 +1,29 @@
-{% macro with_errors(field) %}
-  {% if kwargs.pop('nolabel', '') %}
-    {% set label = False %}
-  {% else %}
-    {% set label = True %}
-  {% endif %}
+{% macro form_field(field) %}
+  {% set css_class = 'form-control ' + kwargs.pop('class', '') %}
   {% if field.errors %}
-    {% set css_class = 'form-control is-invalid ' + kwargs.pop('class', '') %}
-  {% else %}
-    {% set css_class = 'form-control ' + kwargs.pop('class', '') %}
+    {% set css_class = css_class + ' is-invalid ' %}
   {% endif %}
-
-  {% if label %}
+  {% if kwargs.pop('label', True) %}
     <label for="{{ field.name }}" class="mb-0 font-weight-bold">{{ field.label }}</label>
   {% endif %}
   {{ field(class=css_class, **kwargs) }}
-  
   {% if field.errors %}
-  <div class="invalid-feedback">
+    <div class="invalid-feedback">
       {{ field.errors|map('escape')|join(' ') }}
-  </div>
+    </div>
+  {% endif %}
+{% endmacro %}
+
+{% macro with_errors(field) %}
+  {% if field.type == 'FieldList' %}
+    {% if kwargs.pop('label', True) %}
+      <label for="{{ field.name }}" class="mb-0 font-weight-bold">{{ field[0].label }}</label>
+    {% endif %}
+    {% for subfield in field %}
+      {{ form_field(subfield, label=False, **kwargs) }}
+    {% endfor %}
+  {% else %}
+    {{ form_field(field, **kwargs) }}
   {% endif %}
 {% endmacro %}
 
--- a/securitas/templates/user-edit.html
+++ b/securitas/templates/user-edit.html
@@ -35,8 +35,7 @@
             <div class="form-group col">{{ macros.with_errors(form.rhbz_mail) }}</div>
           </div>
           <div class="form-row">
-            {# TODO: We need to have a way to add more keys and fix this styling #}
-            <div class="form-group col">{{ macros.with_errors(form.gpgkeys) }}</div>
+            <div class="form-group col">{{ macros.with_errors(form.gpgkeys, maxlength="16", class="text-monospace mb-1", placeholder="add a GPG Key ID") }}</div>
           </div>
           
           <div class="form-row">
--- a/securitas/controller/user.py
+++ b/securitas/controller/user.py
@@ -56,4 +56,8 @@ def user_edit(ipa, username):
             flash('Profile has been succesfully updated.', 'success')
             return redirect(url_for('user', username=username))
 
+    # Append 2 empty entries at the bottom of the fieldlist
+    for i in range(2):
+        form.gpgkeys.append_entry()
+
     return render_template('user-edit.html', user=user, form=form)
--- a/securitas/form/edit_user.py
+++ b/securitas/form/edit_user.py
@@ -1,7 +1,7 @@
 from flask_wtf import FlaskForm
 from wtforms import FieldList, StringField, SelectField
 from wtforms.fields.html5 import EmailField
-from wtforms.validators import AnyOf, DataRequired, Email, Optional
+from wtforms.validators import AnyOf, DataRequired, Email, Optional, Length
 
 from securitas.utility.locales import LOCALES
 from securitas.utility.timezones import TIMEZONES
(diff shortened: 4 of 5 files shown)


securitas/templates/_form_macros.html
```diff
@@ -1,24 +1,39 @@
...
+      {% set css_class = css_class + ' is-invalid ' %}
+    {% endif %}
+    {% if kwargs.pop('label', True) %}
+      <label for="{{ field.name }}" class="mb-0 font-weight-bold">{{ field.label }}</label>
+    {% endif %}
+    {{ field(class=css_class, **kwargs) }}
+    {% if field.errors %}
+    <div class="invalid-feedback">
+        {{ field.errors|map('escape')|join(' ') }}
+    </div>
```
abompard: I think we can factor some of this code, because I can see some duplication. I suggest the following change:
```diff
--- a/securitas/templates/_form_macros.html
+++ b/securitas/templates/_form_macros.html
@@ -1,6 +1,21 @@
-{% macro with_errors(field) %}
+{% macro form_field(field) %}
   {% set css_class = 'form-control ' + kwargs.pop('class', '') %}
+  {% if field.errors %}
+    {% set css_class = css_class + ' is-invalid ' %}
+  {% endif %}
+  {% if kwargs.pop('label', True) %}
+    <label for="{{ field.name }}" class="mb-0 font-weight-bold">{{ field.label }}</label>
+  {% endif %}
+  {{ field(class=css_class, **kwargs) }}
+  {% if field.errors %}
+  <div class="invalid-feedback">
+      {{ field.errors|map('escape')|join(' ') }}
+  </div>
+  {% endif %}
+{% endmacro %}
+
 
+{% macro with_errors(field) %}
   {% if field.type == 'FieldList' %}
     {# append 4 empty entries at the bottom of the fieldlist #}
     {% for i in range(4) %}
@@ -11,32 +26,14 @@
       <label for="{{ field.name }}" class="mb-0 font-weight-bold">{{ field[0].label }}</label>
     {% endif %}
     {% for subfield in field %}
-      {% if subfield.errors %}
-        {% set css_class = css_class + ' is-invalid ' %}
-      {% endif %}
-      {{ subfield(class=css_class, **kwargs) }}
-      {% if subfield.errors %}
-        <div class="invalid-feedback">
-          {{ subfield.errors|map('escape')|join(' ') }}
-        </div>
-      {% endif %}
+      {{ form_field(subfield, label=False, **kwargs) }}
     {% endfor %}
   {% else %}
-    {% if field.errors %}
-      {% set css_class = css_class + ' is-invalid ' %}
-    {% endif %}
-    {% if kwargs.pop('label', True) %}
-      <label for="{{ field.name }}" class="mb-0 font-weight-bold">{{ field.label }}</label>
-    {% endif %}
-    {{ field(class=css_class, **kwargs) }}
-    {% if field.errors %}
-    <div class="invalid-feedback">
-        {{ field.errors|map('escape')|join(' ') }}
-    </div>
-    {% endif %}
+    {{ form_field(field, **kwargs) }}
   {% endif %}
 {% endmacro %}
 
+
 {% macro non_field_errors(form) %}
   {% if form.errors.non_field_errors %}
     <div id="formerrors">
```

What do you think?

securitas/templates/_form_macros.html
```diff
@@ -1,24 +1,39 @@
...
-  {{ field(class=css_class, **kwargs) }}
-  
-  {% if field.errors %}
-  <div class="invalid-feedback">
-      {{ field.errors|map('escape')|join(' ') }}
-  </div>
+  {% if field.type == 'FieldList' %}
+    {# append 4 empty entries at the bottom of the fieldlist #}
+    {% for i in range(4) %}
+      {% set entry = field.append_entry() %}
```
abompard: If feels a bit strange to be calling it here in this manner, because we don't actually care about the `entry` value we set. On top of that, this will add 4 empty entries for all `FieldList` form fields and we may want to either not do that for other forms or add a different number of empty fields. I think this code should live in the controller. So i suggest removing this here and making this change:
```diff
--- a/securitas/controller/user.py
+++ b/securitas/controller/user.py
@@ -56,4 +56,8 @@ def user_edit(ipa, username):
             flash('Profile has been succesfully updated.', 'success')
             return redirect(url_for('user', username=username))
 
+    # Append 4 empty entries at the bottom of the fieldlist
+    for i in range(4):
+        form.gpgkeys.append_entry()
+
     return render_template('user-edit.html', user=user, form=form)
```

We'll probably have to do the same for SSH keys, but I do think it's better here than in a template (even though Jinja lets us do pretty advanced python inside templates).

securitas/templates/_form_macros.html
```diff
@@ -1,24 +1,39 @@
...
-  {% endif %}
-  {{ field(class=css_class, **kwargs) }}
-  
-  {% if field.errors %}
-  <div class="invalid-feedback">
-      {{ field.errors|map('escape')|join(' ') }}
-  </div>
+  {% if field.type == 'FieldList' %}
+    {# append 4 empty entries at the bottom of the fieldlist #}
+    {% for i in range(4) %}
```
abompard: 4 additional entries seem like a bit too many, no? I would go with 2 (and remember, 640k should be enough for anyone ;-) ).

abompard (changes requested):
It does look good! I have a few comments though, but feel free to disagree with me :-)

securitas/templates/_form_macros.html
ryanlerch: yeah, this is def the better approach. the only reason for `set entry` was because it was returning the feild into the HTML, which was what i didnt want since i was iterating it afterwards.

securitas/templates/_form_macros.html
ryanlerch: agreed. 2 it is

securitas/templates/_form_macros.html
ryanlerch: +1 and updatrred

securitas/templates/user-edit.html
```diff
@@ -36,7 +36,7 @@ <h3 class="header light-blue-text text-darken-4 fw300">
           </div>
           <div class="form-row">
               {# TODO: We need to have a way to add more keys and fix this styling #}
```
abompard: You can probably remove this comment ;-)

securitas/templates/_form_macros.html
```diff
@@ -1,24 +1,30 @@
...
-  <div class="invalid-feedback">
+    <div class="invalid-feedback">
       {{ field.errors|map('escape')|join(' ') }}
-  </div>
+    </div>
+  {% endif %}
+{% endmacro %}
+
+{% macro with_errors(field) %}
+  {% set css_class = 'form-control ' + kwargs.pop('class', '') %}
```
abompard: If you leave that one here, then `class` will have been removed from the `kwargs` and it won't be available in the `form_field` macro. I suggest removing that line since we don't use `css_class` in that block anymore anyway.

abompard (changes requested):

ryanlerch:
@abompard fixed up again, and ready fore re-review

abompard (approved):
Looks great, thanks!

PR #109, opened 2020-02-07
Fix the failing test_group test

ryanlerch:


abompard (approved):
Oh, there's a `select_one()` method! Nice. I should convert the tests that use `select()` and get the first element to that one.

PR #110, opened 2020-02-10
add authed password form

### LINKED ISSUE #44: As a registered user, I want to change my password, so that I can be a responsible person who cares about security.
AC:

A logged-in user can change their password following
They need to provide the current password again for additional security, which is checked against FreeIPA
The password is changed in FreeIPA
They don't need to login again after changing it
A log message is emitted for audit reasons.

DoD: 
all the above are met.

ryanlerch:
Adds a new page for an authenticated user to change
a password. This leaves the unauthenticated form to
just be presented if a user tries to log in with an
expired password.

Resolves: #44

### DIFF
--- a/securitas/controller/password.py
+++ b/securitas/controller/password.py
@@ -1,44 +1,78 @@
-from flask import flash, render_template, redirect, url_for
+from flask import abort, flash, render_template, redirect, request, url_for, session
 import python_freeipa
 
 from securitas import app
-from securitas.security.ipa import untouched_ipa_client
+from securitas.security.ipa import untouched_ipa_client, maybe_ipa_session
+from securitas.utility import with_ipa
 from securitas.form.password_reset import PasswordResetForm
+
+
+def _validate_change_pw_form(form, username, ipa=None):
+    if ipa is None:
+        ipa = untouched_ipa_client(app)
+
+    current_password = form.current_password.data
+    password = form.password.data
+
+    res = None
+    try:
+        res = ipa.change_password(username, password, current_password)
+    except python_freeipa.exceptions.PWChangeInvalidPassword:
+        form.current_password.errors.append(
+            "The old password or username is not correct"
+        )
+    except python_freeipa.exceptions.PWChangePolicyError as e:
+        form.password.errors.append(e.policy_error)
+    except python_freeipa.exceptions.FreeIPAError as e:
+        # If we made it here, we hit something weird not caught above. We didn't
+        # bomb out, but we don't have IPA creds, either.
+        app.logger.error(
+            f'An unhandled error {e.__class__.__name__} happened while reseting '
+            f'the password for user {username}: {e.message}'
+        )
+        form.errors['non_field_errors'] = ['Could not change password.']
+
+    if res and res.ok:
+        flash('Your password has been changed', 'success')
+        app.logger.info(f'Password for {username} was changed')
+    return res
 
 
 @app.route('/password-reset', methods=['GET', 'POST'])
 def password_reset():
+    # If already logged in, redirect to the logged in reset form
+    ipa = maybe_ipa_session(app, session)
+    username = session.get('securitas_username')
+    if ipa and username:
+        return redirect(url_for('auth_password_reset', username=username))
+
+    username = request.args.get('username')
+    if not username:
+        abort(404)
     form = PasswordResetForm()
 
     if form.validate_on_submit():
-        username = form.username.data
-        current_password = form.current_password.data
-        password = form.password.data
-
-        ipa = untouched_ipa_client(app)
-        res = None
-        try:
-            res = ipa.change_password(username, password, current_password)
-        except python_freeipa.exceptions.PWChangeInvalidPassword:
-            form.current_password.errors.append(
-                "The old password or username is not correct"
-            )
-        except python_freeipa.exceptions.PWChangePolicyError as e:
-            form.password.errors.append(e.policy_error)
-        except python_freeipa.exceptions.FreeIPAError as e:
-            # If we made it here, we hit something weird not caught above. We didn't
-            # bomb out, but we don't have IPA creds, either.
-            app.logger.error(
-                f'An unhandled error {e.__class__.__name__} happened while reseting '
-                f'the password for user {username}: {e.message}'
-            )
-            form.errors['non_field_errors'] = ['Could not change password.']
-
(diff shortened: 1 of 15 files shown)


ryanlerch:
@abompard @StephenCoady this one is ready for review now

securitas/controller/password.py
```diff
@@ -1,44 +1,75 @@
...
+def auth_password_reset(ipa, username):
+    if session.get('securitas_username') != username:
+        flash('You do not have permission to edit this account.', 'danger')
+        return redirect(url_for('root'))
 
+    form = PasswordResetForm()
+
+    if form.validate_on_submit():
+        res = _validate_change_pw_form(form, ipa)
+        app.logger.error(res)
```
abompard: This looks like a leftover, you'd log it even if there's no error.

securitas/security/ipa.py
```diff
@@ -11,7 +11,10 @@ def parse_group_management_error(data):
 
     TODO: send this upstream.
     """
-    failed = data['failed']
+    try:
+        failed = data['failed']
+    except KeyError:
+        return
```
abompard, line 17: This new branch will probably need a unit test, sorry... ;-)

securitas/controller/password.py
```diff
@@ -1,44 +1,75 @@
...
+        # bomb out, but we don't have IPA creds, either.
+        app.logger.error(
+            f'An unhandled error {e.__class__.__name__} happened while reseting '
+            f'the password for user {username}: {e.message}'
+        )
+        form.errors['non_field_errors'] = ['Could not change password.']
+
+    if res and res.ok:
+        flash('Your password has been changed', 'success')
+        app.logger.info(f'Password for {username} was changed')
```
abompard, line 37: I think it would be useful to also log something when the call was unsuccessful.

abompard (changes requested):

securitas/tests/unit/controller/test_password.py
```diff
@@ -16,6 +18,81 @@ def test_password_reset(client):
     assert page.title.string == 'Password Reset - The Fedora Project'
 
 
+@pytest.mark.vcr()
+def test_password_reset_user(client, logged_in_dummy_user):
+    """Test the redirect to the authed password reset"""
+    result = client.get('/password-reset')
+    assert result.location == f"http://localhost/user/dummy/password-reset"
```
StephenCoady, line 29: might be a silly question but will this always be the case? wondering will localhost be true across environments.

abompard: As long as it's run from the Flask test runner (the `client` instance here) it should always be localhost.

securitas/security/ipa.py
ryanlerch: added a test for this one.

securitas/controller/password.py
ryanlerch: is the logging we arfe doing in _validate_change_pw_form() not enough here? it logs if we hit some errors.

Not sure how to avoid logging all validation errors here, as these will show up here too, right?

securitas/controller/password.py
ryanlerch: removed.

ryanlerch:
Everything fixed, with one caveat, which is unresolved, and commented on above.

Coverage for this PR is now back up to 100% too.

securitas/controller/password.py
abompard: Yeah you're right, it's fine.

abompard (commented):
OK that part is good!

I just thought of something, though: the only way to end up on the password expired page now, is to be redirected there from the login form when the password is expired. In that situation, we do know what the username is also: it's the one that the user just provided.
As a result I think we can remove the username input field from the non-authed password reset page as well. What do you think?

ryanlerch:
@abompard added the extra functionality to the unaiuthed form.  Kept in a seperate commit for ease of review. will squash when approved.

securitas/templates/password-reset.html
```diff
@@ -13,16 +13,14 @@
...
+        <form action="{{ url_for('password_reset', username=username) }}" method="post">
         {% endif %}
           <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
           {% if current_user %}
           <input type="hidden" name="username" value="{{current_user.username}}"/>
           {% else %}
-          <div class="form-row">
-            <div class="form-group col">{{ macros.with_errors(password_reset_form.username, tabindex="1")}}</div>
-          </div>
+          <input type="hidden" name="username" value="{{username}}"/>
```
abompard: If you remove the field from the form you don't need this.

securitas/tests/unit/controller/test_password.py
```diff
@@ -121,7 +122,7 @@ def test_password_user(client, logged_in_dummy_user):
 def test_non_matching_passwords(client):
     """Verify that passwords that dont match are caught"""
     result = client.post(
-        '/password-reset',
+        '/password-reset?username=jbloggs',
         data={
             "username": "jbloggs",
```
abompard: If you remove the field from the form you'll have to remove all those too

abompard (changes requested):
Also please test the case where the user passed in the query string does not exist in IPA

ryanlerch:
ok. updated again.

securitas/tests/unit/controller/test_password.py
```diff
@@ -50,6 +112,48 @@ def test_password(client, dummy_user):
...
+    assert result.status_code == 200
+    page = BeautifulSoup(result.data, 'html.parser')
+    password_input = page.select("input[name='password']")[0]
+    assert 'is-invalid' in password_input['class']
+    invalidfeedback = password_input.find_next('div', class_='invalid-feedback')
+    assert invalidfeedback.get_text(strip=True) == "Passwords must match"
+
+
+@pytest.mark.vcr()
+def test_password(client, dummy_user):
```
abompard, line 144: Could you also add a test like this one but without the `dummy_user` fixture, so that we test the case where the username does not exist? I don't think we need to display a different user message but I'd prefer it tested.

abompard (approved):
Yeah, looks good to me. I've added a minor suggestion, but feel free to ignore it if you want to, it's not a blocker to me.

securitas/tests/unit/controller/test_password.py
ryanlerch: done!

PR #116, opened 2020-02-13
remove unused wtforms.StringField import in password-reset.py

ryanlerch:


abompard:
I did this in another commit that's already merged.

PR #120, opened 2020-02-17
Add the ability to add and remove SSH keys

### LINKED ISSUE #40: As a registered user, I want to be able to add and remove SSH keys from my account, so that I can access Fedora systems with my current keys and disable access from old keys.

ryanlerch:
Resolves: #40

### DIFF
--- a/securitas/controller/user.py
+++ b/securitas/controller/user.py
@@ -45,6 +45,7 @@ def user_edit(ipa, username):
                 full_name='%s %s' % (form.firstname.data, form.lastname.data),
                 display_name='%s %s' % (form.firstname.data, form.lastname.data),
                 mail=form.mail.data,
+                ipasshpubkey=form.sshpubkeys.data,
                 fasircnick=form.ircnick.data,
                 faslocale=form.locale.data,
                 fastimezone=form.timezone.data,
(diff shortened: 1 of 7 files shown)


securitas/controller/user.py
```diff
@@ -68,5 +69,6 @@ def user_edit(ipa, username):
     # Append 2 empty entries at the bottom of the fieldlist
     for i in range(2):
```
StephenCoady: Is this correct? If its intentional it might be confusing to have two empty fields originally. It also means if you actually have two keys you get 4 boxes with two being empty.
Would a better default be only show one empty box for each key and when one is added its number of keys + one empty field?

StephenCoady:
I also tried to add the same key twice. The app correctly didn't let me but I got no banner telling me why it wasn't added - I was just booted back to my profile page.

securitas/controller/user.py
ryanlerch: yeah, for gpgkeys, we decided on 2 empty boxes. Probably for the SSH keys, one will work

ryanlerch:
> I also tried to add the same key twice. The app correctly didn't let me but I got no banner telling me why it wasn't added - I was just booted back to my profile page.

Yeah, freeIPA simply only adds one key if duplicates are provided. However the issue you are running into here is that because the first key is already saved, the duplicate one you are adding is ignored, so there is no change to the user. 

I have updated the code here In the case of submittting the form with no change, to stay on the form, and return the "no modifications to be performed" error to the form.

StephenCoady (approved):

PR #126, opened 2020-02-18
fix bugs found by sonarqube

### LINKED ISSUE #101: As a developer I want a secure coding tool to analyse my code so that it adheres to best practices
https://www.sonarqube.org/ is a code analyser which can identify bugs in code and also highlight best practices from a secure coding point of view. I think it would be good to have this tool as part of this project, especially considering it is an identity management solution.

In my opinion this user story could be broken into multiple parts.

1) Analise securitas using all 3 tools mentioned in this issue
2) Provide manual feedback to the team from this first report
3) Fix any and all issues the tools identify
4) Investigate the 3 tools and where/how we could host it or have it hosted
5) Investigate how these services could be integrated into our CI pipeline

I think 4 and 5 are stretch goals and will probably require input from someone with more operations knowledge than myself.

**AC**
1. We have generated enough information to make an informed decision about which of these tools we will use.
1. We have information about how the tool(s) we select could become integrated into our workflow.
1. Any issues identified by these tools are fixed.

**DoD**

This issue can be considered done when this repo has had any and all security issues fixed and at least one of the tools investigated is part of the development workflow.

james02135:
this commit fixes errors found by sonarqube and implements best practices

Signed-off-by: james02135 <james02135@hotmail.com>


securitas/themes/fas/static/fonts/font-awesome.css
```diff
@@ -14,6 +14,7 @@
 .fa {
   display: inline-block;
   font: normal normal normal 14px/1 FontAwesome;
+  font-family: Arial, Helvetica, sans-serif;
```
relrod: What is this line doing? This is for an icon font, it's none of those families, right?

securitas/themes/fas/static/js/fedora-bootstrap/fedora-bootstrap.css
```diff
@@ -51,22 +51,18 @@ article, aside, figcaption, figure, footer, header, hgroup, main, nav, section {
 
 body {
   margin: 0;
-  font-family: "Open Sans";
+  font-family: Arial, Helvetica, sans-serif;
```
relrod: Why this change?

securitas/themes/fas/static/js/fedora-bootstrap/fedora-bootstrap.css
```diff
@@ -7317,7 +7307,7 @@ body {
   background: #f7f7f9; }
 
 .c-widget-header h6 {
-  font-family: "Open Sans";
+  font-family: Arial, Helvetica, sans-serif;
```
relrod: Why?

securitas/themes/fas/static/js/fedora-bootstrap/fedora-bootstrap.css
```diff
@@ -7358,7 +7348,7 @@ body {
   padding: 15px 20px; }
   .c-widget-meeting-event h6, .c-widget-meeting-event h5, .c-widget-meeting-event p {
     color: #55595c;
-    font-family: "Open Sans"; }
+    font-family: Arial, Helvetica, sans-serif; }
```
relrod: Why?

securitas/themes/fas/static/js/fedora-bootstrap/fedora-bootstrap.css
```diff
@@ -7386,7 +7376,7 @@ body {
   padding: 15px 20px; }
   .c-widget-meeting-request h6, .c-widget-meeting-request h5 {
     color: #55595c;
-    font-family: "Open Sans"; }
+    font-family: Arial, Helvetica, sans-serif; }
```
relrod: Why?

securitas/templates/group.html
```diff
@@ -86,7 +86,7 @@ <h4>Sponsors</h4>
                 <div class="text-muted">
                   {% if current_user_is_sponsor %}
                     <button type="submit" class="btn btn-outline-danger border-0" name="username" value="{{ member.username }}">
-                      <i class="fa fa-trash"></i>
+                      <em class="fa fa-trash"></em>
```
relrod: `<em>` is not the right replacement here. If you hate `<i>` for some reason, then `<span>` is the correct replacement here.

relrod (commented):
I'm -1 on this change and stopped my review half way through.
These are more changes that, like Black in my opinion, change things for no practical gain.
Some of these changes don't even make sense, it's adding an Arial font-family to an icon font? It's replacing `<i>` with `<em>` for an icon font? It's using `<strong>` instead of `<b>` when there's no semantic reason for these labels to have [strong importance](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong)? Why??

securitas/themes/fas/static/fonts/font-awesome.css
ryanlerch: Additionally, this basically an bundled library. so im -1 for changing this file

securitas/themes/fas/static/js/fedora-bootstrap/fedora-bootstrap.css
ryanlerch: same thing, this basically an bundled library. so i'm -1 for changing this file

StephenCoady (commented):
This looks good to me. I won't approve though as @abompard will be the one who needs to merge when the org is ready on sonarcloud. Great job!

StephenCoady:
@james02135 you do need to ammend your commit with your sign-off though.

sonar-project.properties
```diff
@@ -0,0 +1,10 @@
+sonar-project.properties
```
abompard, line 1: I don't think this line is required, I haven't found it in the config file examples or in the configuration parameters documentation.

sonar-project.properties
```diff
@@ -0,0 +1,10 @@
...
+sonar.organization=fedora-infra
+sonar.projectKey=fedora-infra/noggin
+sonar.sources=.
```
abompard: shouldn't this be `noggin`?

sonar-project.properties
```diff
@@ -0,0 +1,10 @@
...
+sonar.host.url=https://sonarcloud.io
+sonar.login=a36fe78f64abe3b8e6db0df06a6b399c2abc2e24
+sonar.test.exclusions=**/noggin/tests/*
```
abompard, line 7: I'm not sure why the double star at the beginning. The tests are under `noggin/tests` so this should probably be `noggin/tests/**`, no?
Also, are you sure it's `sonar.test.exclusions` and not `sonar.exclusions`? I haven't found a clear distinction between the two in [the doc](https://docs.sonarqube.org/latest/project-administration/narrowing-the-focus/).

sonar-project.properties
```diff
@@ -0,0 +1,10 @@
...
+sonar.exclusions=/devel/create-test-data.py, /noggin/themes/*, /noggin/templates/*,
```
abompard: Not sure that paths should begin with a `/`. Also, why should we exclude those paths?

sonar-project.properties
```diff
@@ -0,0 +1,10 @@
+sonar-project.properties
+sonar.organization=fedora-infra
+sonar.projectKey=fedora-infra/noggin
+sonar.sources=.
+sonar.host.url=https://sonarcloud.io
+sonar.login=a36fe78f64abe3b8e6db0df06a6b399c2abc2e24
```
abompard, line 6: This is not a private token, is it?

abompard (changes requested):

sonar-project.properties
james02135: I tried to run sonarqube without that line after SCoady suggested the same, and it wouldn't run so I left it in.

sonar-project.properties
james02135: It is, will need to be changed when the webhook is implemented

sonar-project.properties
james02135: There is a difference between "test" exclusions and "regular" exclusions, I ran this file a few different ways to determine how to write the exclusions properly.

sonar-project.properties
james02135: The create-test-data file is just a way for us to create users and groups so that there is something in "noggin", not something we need sonarqube to test against. RLerch let me know that the other two paths, "themes" and "templates" are where the bundled libraries are kept and also shouldn't be tested against.

sonar-project.properties
james02135: I'm sorry, I'm not sure what you mean by "this"

abompard: The `sonar.sources` value should point at the directory where the sources are. This should probably be `noggin` since that's where our code lives, I don't think we need Sonarqube to analyze the ansible files, default config files, etc. No?

sonar-project.properties
abompard: The themes also contain templates and I thought we wanted to test them, for example that's where the fixes on the `alt` tag for images came from. If we want to keep templates but drop bundled libraries, maybe we could use something like `noggin/themes/*/static/`?

sonar-project.properties
abompard: OK, so when the webhook is added this line can be removed?

sonar-project.properties
james02135: You're right, that's changed to noggin now

sonar-project.properties
james02135: that has been changed

sonar-project.properties
james02135: It will definitely need to be changed at least, I'm not positive if you will need to add your own token or not.

james02135:
@abompard, sonarcloud has been added to the automated tests for noggin so unless you have any changes you'd like me to do on the PR, it should be ready to merge

abompard (approved):
Yes, I can see the checks! Thanks.

PR #128, opened 2020-02-19
Add the S2I files to build in OpenShift

abompard:
This adds the Source To Image files that OpenShift needs to build and run Securitas.


abompard:
I'm just going to merge this because there's little to review, and it's only a small part of enabling an OpenShift deployment.

PR #131, opened 2020-02-21
check CI

ryanlerch:


abompard:
I found why it was failing on rawhide, I think you can close this.

PR #133, opened 2020-02-21
Add openSUSE theme

hellcp:


securitas/templates/master.html
```diff
@@ -6,7 +6,7 @@
     {% block head %}
     {% endblock %}
     <link href="{{ url_for('static', filename='css/main.css') }}" rel="stylesheet" type="text/css">
-    <title>{% block title %}{% endblock %} - {{ project }}</title>
+    <title>{% block title %}{% endblock %} - {% block website %}securtias{% endblock %}</title>
```
Conan-Kudo: Typo here...

securitas/themes/openSUSE/static/css/accounts.css
```diff
@@ -0,0 +1,19 @@
+/*CSS rules for the 'FAS' securitas theme */
```
Conan-Kudo: Isn't this the `openSUSE` one?

Conan-Kudo (changes requested):

Conan-Kudo:
@hellcp Could you squash the fixups back into the main commit? Looks good otherwise!

Conan-Kudo:
Anyone know what this is about?

```
Command ['/workdir/workspace/f31/.tox/licenses/bin/pip', 'install', '--no-deps', 'entrypoints==0.3'] errored with the following return code 1, and output: 

Collecting entrypoints==0.3

  ERROR: Could not find a version that satisfies the requirement entrypoints==0.3 (from versions: none)

```

securitas/app.py
```diff
@@ -14,7 +14,6 @@
 def inject_global_template_vars():
     # TODO: move project out to config var
     return dict(
-        project="The Fedora Project",
```
ryanlerch, line 17: +1
Thanks for moving this out of the code and into the templates / themes

securitas/themes/default/static/css/default.css
```diff
@@ -1 +1,9 @@
 /*CSS rules for the 'default' securitas theme */
+
+html, body {
+  height: 100vh;
+  max-width: 100vw;
+  margin: 0;
+  display: flex;
+  flex-flow: column;
+}
```
ryanlerch: What are these rules fixing here? Is this something the stock bootstrap doesnt give us?

ryanlerch (commented):
Also, it might be good to split this up into two dinstinct commits, one that does the changes to the main templates, and the current themes. Then another that adds the openSUSE theme.

securitas/themes/default/static/css/default.css
hellcp: Eh, frankly this would work as well in the shared css file, it's only here because this bit is different for the Fedora theme, but it shouldn't matter in the end

StephenCoady (approved):

PR #134, opened 2020-02-21
Use libravatar with the Gravatar-compatible API endpoint

Conan-Kudo:
The libravatar service provides a FOSS drop-in replacement to
Gravatar, and is already in use by many other Fedora services.
Thus, we should be use it to provide consistency in avatar
representation.

Signed-off-by: Neal Gompa <ngompa13@gmail.com>


hellcp:
As an additional point, it would be nice to provide a link in user edit linking to libravatar like badges.fedoraproject.org already does.

StephenCoady (approved):

PR #136, opened 2020-02-22
Make the templates a little nicer

hellcp:
**Combined login and register dialog on the index page**
Index
Index(1)

**Narrower profile for easier consumption of information**
Profile
Profile(1)

**Settings split into 3 parts (and password change integrated into them)**
Settings
Settings(1)
Settings(2)

**Group view with a little space saving applied**
Group

**Narrower Group List**
Group List

**The rest of the views got cards now**
Screenshot_2020-02-22 Password Recovery - The Fedora Project

Outside of that, title fixes so they are more compact and a little more human parsable.

Signed-off-by: Stasiek Michalski <stasiek@michalski.cc>

### DIFF
--- a/securitas/controller/user.py
+++ b/securitas/controller/user.py
@@ -6,6 +6,7 @@ from securitas.form.edit_user import EditUserForm
 from securitas.representation.group import Group
 from securitas.representation.user import User
 from securitas.utility import with_ipa, user_or_404
+from securitas.form.password_reset import PasswordResetForm
 
 
 @app.route('/user/<username>/')
@@ -35,6 +36,7 @@ def user_edit(ipa, username):
 
     user = User(user_or_404(ipa, username))
     form = EditUserForm(obj=user)
+    password_reset_form = PasswordResetForm()
 
     if form.validate_on_submit():
         try:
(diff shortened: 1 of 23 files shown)


ryanlerch:
Overall, +1 to the changes.

With the updated user profile page, not sure spliting the groups into a seperate tab is worth it here. What is the rationale for obscuring this behind another tab?

hellcp:
The columned profile didn't work well on smaller screen sizes, and to make the interface switch between this and columned profile I would need to pretty much duplicate the entire template in one view and switch between them on the fly, making the template quite a big bigger and slower to load.

ryanlerch:
From a UI point of view, i'm +1 to this. Might get one of the other devs to do a pass over this too.

Two notes (take em or leave em)
* Did the tab style look ok for the new sub-pages navs? IMHO, they lend more to being tabs. You are using tabs on the frontpage for register / login.
* I *think* we are going to have one more field in the register form, which might make that box a little larger. Would it work if the welcome jumbo is at the top, and login / register is below. It is moving the main point of the page possibly below the fold, but does remove a col layout all together.

ryanlerch (approved):

hellcp:
I avoided using tabs for the settings because tabs look terrible when they are split into 2 lines when the container runs out of space. In case there would be more options in the card on the home page, I would probably switch to pills too (except, I don't really see how you could have the third option to login/register, so the tabs made sense).

For the homepage jumbo login under the text, I am not opposed to it, I did find that side-by-side isn't the best layout option for it, but it is fine for now, considering there isn't that much content there anyway right now. You can always link to scroll down to the login/register box when it goes below the fold.

PR #138, opened 2020-02-24
Add avatar editing to the settings

hellcp:
Kinda requires #134 considering it uses libravatar for the settings

### DIFF
--- a/securitas/templates/user-edit.html
+++ b/securitas/templates/user-edit.html
@@ -24,21 +24,15 @@
             <div class="tab-pane fade show active" id="settings-personal" role="tabpanel">
               <div class="card-body pb-0">
                 <label>Avatar</label>
-                <div class="row">
-                  <div class="col-auto">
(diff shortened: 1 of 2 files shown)


hellcp:
image

StephenCoady:
Hi @hellcp thanks for this! I do wonder though, do we need the change avatar button really? and if we want to link to libravatar then maybe a line under the avatar with a small link? 

It just seems a very big button to land so forefront on the screen to me for something that most people may not care too much about.

securitas/templates/user-edit.html
```diff
@@ -22,6 +22,18 @@
           </ul>
           <div class="tab-content">
             <div class="tab-pane fade show active" id="settings-personal" role="tabpanel">
+              <div class="card-body pb-0">
+                <label>Avatar</label>
```
StephenCoady: I think the formatting here is a bit funny now. We could probably just remove the word avatar and add it to alt of the img just in case. Also could we put the change avatar link below the actual avatar?

StephenCoady (approved):

PR #139, opened 2020-02-25
Add email field to register form

### LINKED ISSUE #127: As a User, i want to supply my email address when creating a new account
Currently, when creating a new account, there is no option for adding an email address. It can be updated later, but it would be good if it was added in the initial form.

I suppose it should be required too, since it is required when editing a user too...

ryanlerch:
Previously, the email field was not present on the
register form. This commit adds the field.

Resolves: #127


StephenCoady (approved):
Verified everything working as expected, all tests passing and coverage is still chockers at 100% :smiley:

PR #146, opened 2020-02-27
fedora-messaging integration

### LINKED ISSUE #140: As an application owner, I want a fedora-messaging message to be emitted when a user is sponsored into a group
We need the portal to emit a message when a member is sponsored into a group, because the Badges application uses it. FAS used to emit a message on the `fas.group.member.sponsor` topic, the contents could be modelled after it (or adapted).

**Acceptance Criteria**:
- a Fedora Messaging message is emitted when a user is sponsored into a group
- to ease transition and if it's possible, the message should be emitted on the same topic with the same content.
- a crash in the message sending code should cause a message to be logged with full traceback, but the page should not fail rendering.

**Definition of Done**:
- all the above is done
- unit tests for 100% coverage (unless explicitely ignored with an explanation)
- CI passes
- the path to the Fedora Messaging configuration can be set by a configuration variable.
- the message should use [schemas](https://fedora-messaging.readthedocs.io/en/stable/messages.html)
- there is documentation on the message schema and its contents

ryanlerch:
This adds support for fedora messaging, by adding a few things.

* Set up RabbitMQ server in the vagrant box, so we can test
  when messages are sent. During development, have the command
  `fedora-messaging consume` running and messages sent will
  appear in your terminal.
* add fedora-messaging as a dep
* add the first message schema for fas.group.member.sponsor which
  is the message that is emmited when a sponsor adds a user to a group.
  Note that the current format of this message is:
  https://apps.fedoraproject.org/datagrepper/id?id=2019-6e4ddbd8-b51b-4753-a4f3-6106ccd560e8&is_raw=true&size=extra-large
* use backoff to retry sending the message, and when finally giving up,
  log the exception.

Note, not 100% Coverage yet with tests, wanted to make sure i was doing this right before rwiting new tests.

Resolves: #140


securitas/controller/group.py
```diff
@@ -62,6 +63,11 @@ def group_add_member(ipa, groupname):
             return redirect(url_for('group', groupname=groupname))
 
         flash('You got it! %s has been added to %s.' % (username, groupname), 'success')
+
+        config.conf.setup_logging()
+        msg = schema.MemberSponsorV1()
```
ryanlerch: When trying to use my schema here, this fails with the message:

    TypeError: The class <class 'securitas.messaging.schema.MemberSponsorV1'> is not in the message registry, which indicates it is not in the current list of entry points for "fedora_messaging". Please check that the class has been added to your package's entry points.

This bit i added in the pyproject.toml was me trying to add it to the entry points, but it doenst seem to be wokring:

    [tool.poetry.plugins."fedora.messages"]
    "fas.group.member.sponsor.v1"="securitas.messaging.schema:MemberSponsorV1"

securitas/controller/group.py
```diff
@@ -62,6 +63,11 @@ def group_add_member(ipa, groupname):
...
+        api.publish(api.Message(topic="hello", body={"Hello": "world!"}))
```
ryanlerch: if you comment out using the schema above, this message goes through to the queue when adding a new user to a group.

to view this message when  it comes in, use `fedora_messaging consume` and have ti running when messages are sent and they will print on the terminal for you

securitas/messaging/schema.py
```diff
@@ -0,0 +1,34 @@
+from fedora_messaging import message
+
+class MemberSponsorV1(message.Message):
```
ryanlerch: I think this is right, just cant test it because of the entry points issue above

devel/ansible/roles/dev/defaults/main.yml
```diff
@@ -0,0 +1 @@
+rabbitmq_cluster_file_limit: 500000
```
ryanlerch, line 1: all these changes to the vagrant box are to install rabbitmq.

devel/ansible/roles/dev/tasks/rabbitmq.yml
```diff
@@ -0,0 +1,30 @@
...
+      state: present
+  with_items:
+      - rabbitmq-server
+
+- name: Create RabbitMQ systemd override directory
+  file:
+    path: /etc/systemd/system/rabbitmq-server.service.d/
+    state: directory
+
+- name: Override rabbitmq hostname
```
ryanlerch, line 14: without this, we get an error like

    Feb 27 08:36:05 ipa.example.com rabbitmq-server[13914]: ERROR: epmd error for host ipa: address (cannot connect to host/port)

when trying to start rabbitmq

pyproject.toml
```diff
@@ -41,6 +42,9 @@ black = {version = "^18.3-alpha.0", allow-prereleases = true}
 flake8 = "^3.7"
 Faker = "^4.0.0"
 
+[tool.poetry.plugins."fedora.messages"]
```
ryanlerch: trying to add entry_points for the schema here. does it work? nope.

abompard:
Here is the test file that I made:
```python
from fedora_messaging import api, config
from securitas.messaging import schema

config.conf.setup_logging()
msg = schema.MemberSponsorV1()
api.publish(api.Message(topic="hello", body={"Hello": "world!"}))
```

Now when I run:
```
$ poetry run python testmsg.py 
[INFO fedora_messaging.message] Registering the 'fas.group.member.sponsor.v1' key as the '<class 'securitas.messaging.schema.MemberSponsorV1'>' class in the Message class registry
[INFO fedora_messaging.message] Registering the 'base.message' key as the '<class 'fedora_messaging.message.Message'>' class in the Message class registry
[INFO twisted] Log opened.
[INFO twisted] Starting factory FedoraMessagingFactoryV2(parameters=<URLParameters host=localhost port=5672 virtual_host=/ ssl=False>, confirms=True)
[INFO fedora_messaging.twisted.protocol] Waiting for 0 consumer(s) to finish processing before halting
[INFO fedora_messaging.twisted.protocol] Finished canceling 0 consumers
[INFO fedora_messaging.twisted.protocol] Disconnect requested, but AMQP connection already gone
```

And if I have a `fedora-messaging consume` command opened on the side it does get the message. It also seems to work when I use the message class you wrote (and I add the `msg` property which is set as required).

So, something else in your setup is breaking things maybe?

ryanlerch:
> Here is the test file that I made:
> 
> ```python
> from fedora_messaging import api, config
> from securitas.messaging import schema
> 
> config.conf.setup_logging()
> msg = schema.MemberSponsorV1()
> api.publish(api.Message(topic="hello", body={"Hello": "world!"}))
> ```
> 
> Now when I run:
> 
> ```
> $ poetry run python testmsg.py 
> [INFO fedora_messaging.message] Registering the 'fas.group.member.sponsor.v1' key as the '<class 'securitas.messaging.schema.MemberSponsorV1'>' class in the Message class registry
> [INFO fedora_messaging.message] Registering the 'base.message' key as the '<class 'fedora_messaging.message.Message'>' class in the Message class registry
> [INFO twisted] Log opened.
> [INFO twisted] Starting factory FedoraMessagingFactoryV2(parameters=<URLParameters host=localhost port=5672 virtual_host=/ ssl=False>, confirms=True)
> [INFO fedora_messaging.twisted.protocol] Waiting for 0 consumer(s) to finish processing before halting
> [INFO fedora_messaging.twisted.protocol] Finished canceling 0 consumers
> [INFO fedora_messaging.twisted.protocol] Disconnect requested, but AMQP connection already gone
> ```
> 
> And if I have a `fedora-messaging consume` command opened on the side it does get the message. It also seems to work when I use the message class you wrote (and I add the `msg` property which is set as required).
> 
> So, something else in your setup is breaking things maybe?

Hmm, when i run that script it seems to work now too... just when running it from within securitas it seems to fail...

abompard:
I tied sponsoring a member into a group from the UI, and it did send the message.

devel/ansible/roles/dev/tasks/rabbitmq.yml
abompard, line 14: It worked without this step for me. Can you try again and if it fails again maybe we can debug it together?

securitas/utility/messaging.py
```diff
@@ -0,0 +1,26 @@
...
+    (fml_exceptions.ConnectionException, fml_exceptions.PublishReturned),
+    max_tries=3,
+    on_backoff=backoff_hdlr)
+def _publish(message):
+    api.publish(message)
+
+def publish(message):
+    try:
+        _publish(message)
+    except (fml_exceptions.ConnectionException, fml_exceptions.PublishReturned) as e:
```
abompard: Here I think you can simply catch `fml_exceptions.BaseException`, because you want the message to be logged whether it's a temporary error or not.

securitas/utility/messaging.py
```diff
@@ -0,0 +1,26 @@
+import sys
+import traceback
+
+from fedora_messaging import api, exceptions as fml_exceptions
+import backoff
+
+from securitas import app
+
+def backoff_hdlr(details):
+    app.logger.error(f"Publishing message failed. Retrying. {traceback.format_tb(sys.exc_info()[2])}")
```
abompard: I think this can be a warning-level log.

securitas/utility/messaging.py
```diff
@@ -0,0 +1,26 @@
+import sys
+import traceback
+
+from fedora_messaging import api, exceptions as fml_exceptions
+import backoff
+
+from securitas import app
```
abompard, line 6: I would recommend using `from flask import current_app` instead, it can help solve circular imports.

securitas/utility/messaging.py
```diff
@@ -0,0 +1,26 @@
+import sys
+import traceback
+
+from fedora_messaging import api, exceptions as fml_exceptions
+import backoff
```
abompard, line 4: The `import X` lines usually go before the `from X import Y` lines.

securitas/utility/messaging.py
```diff
@@ -0,0 +1,26 @@
...
+    max_tries=3,
+    on_backoff=backoff_hdlr)
+def _publish(message):
+    api.publish(message)
+
+def publish(message):
+    try:
+        _publish(message)
+    except (fml_exceptions.ConnectionException, fml_exceptions.PublishReturned) as e:
+        app.logger.error(f"Publishing message failed. Giving up. {traceback.format_tb(sys.exc_info()[2])}")
```
abompard: I think you can use `app.logger.exception`, it'll log the traceback automatically.

securitas/utility/messaging.py
```diff
@@ -0,0 +1,26 @@
...
+import backoff
+
+from securitas import app
+
+def backoff_hdlr(details):
+    app.logger.error(f"Publishing message failed. Retrying. {traceback.format_tb(sys.exc_info()[2])}")
+
+@backoff.on_exception(
+    backoff.expo,
+    (fml_exceptions.ConnectionException, fml_exceptions.PublishReturned),
```
abompard: Please replace `PublishReturned` with `PublishException` (see [bodhi commit a3ff73](https://github.com/fedora-infra/bodhi/commit/a3ff73d0c57236cd1d9d38d967c029b7d2b6f435))

pyproject.toml
```diff
@@ -41,6 +43,9 @@ black = {version = "^18.3-alpha.0", allow-prereleases = true}
...
+"fas.group.member.sponsor.v1"="securitas.messaging.schema:MemberSponsorV1"
```
abompard: I would recommend prefixing with `securitas` as it's not related to the topic but to the app.

abompard (changes requested):
The main change that is also necessary is that message schemas actually need to be in their own python packages, so apps that listen to them can install the package and not the whole of securitas...
So I would recommend actually making a separate repo in Github called securitas-messages (to be renamed later :'-(  ), with its own `pyproject.toml` file, etc, and add it to the securitas dependencies.
I'm going to jumpstart that repo for you.

Here it is: https://github.com/fedora-infra/securitas-messages. You should have write access to it.

ryanlerch:
Okay, moved the schema to the separate repo, and added it as a git dep in poetry.

Also have added tests to keep the coverage at 100%.

ryanlerch:
@abompard any thoughts on why this is failing liccheck? is it because i am pulling in securitas-messages with git?

abompard:
I think you need to add `GPLv2+` to the allowed licenses in `pyproject.toml`. It also looks like liccheck can't find the license for `securitas-messages`.

abompard (approved):
Yeah, I think it's good now :-)

PR #157, opened 2020-03-10
User disable

### LINKED ISSUE #43: As a registered user, I want to temporarily disable my account, so that I can come back later if I decide to.
Acceptance Criteria:
Spike what happens when a user is disabled and they try to login
DOD
Above is complete

StephenCoady:
this pr adds the ability for users to disable their accounts. there has been a restructure of the front end since i started this so i'll need to fix that first.

resolves #43


StephenCoady:
I'm not really sure why the coverage is being reported as 99% as I've double checked and it is definitely hitting the lines it thinks it isn't. I'll need to look into it

PR #158, opened 2020-03-11
Add the ability to add a new OTP token

### LINKED ISSUE #147: As a User, I want to be able to enroll an OTP token, so that I use 2FA and improve security
**Acceptance Criteria**
- enrollment of a token should be done after release.
- the page should display a QR Code so users can easily add the token to their favorite app
- the page should refuse enrolling a token if there is no GPG key present, otherwise there would be no way to recover the account if the token is lost.
- the page should check the password again before enrolling the token

**Definition of Done**
- All of the above works
- Documentation is written
- CI passes

ryanlerch:
This adds the ability in securitas to create a new
OTP token, as well as listing the tokens in the settings
UI.

Resolves: #147


StephenCoady (approved):
This looks good to me, I can't see anything wrong with it and i've tested everything is working. I know you said there is just the issue of the password but I'm happy enough for us to create a follow on issue to protect them again with entering the password or something.

PR #162, opened 2020-03-13
add disable option for otp tokens

### LINKED ISSUE #148: As a User, I want to be able to disable an OTP token
#147 should be completed first 

**Acceptance Criteria**
- the page should be authenticated
- it should not be possible to disable the last token (as FreeIPA will [refuse the request](https://www.freeipa.org/page/V4/OTP#User-Managed_Tokens))

**Definition of Done**
- All of the above works
- Documentation is written
- CI passes

StephenCoady:

resolves #148

### DIFF
--- a/noggin/controller/user.py
+++ b/noggin/controller/user.py
@@ -153,7 +153,7 @@ def user_settings_otp(ipa, username):
         select="otp",
         tokens=tokens,
         otp_uri=otp_uri,
-        enabled_count=enabled_count
+        enabled_count=enabled_count,
     )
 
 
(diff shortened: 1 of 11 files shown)


ryanlerch:
Looks like the tests are just failing on style.

a `poetry run black .` should fix these issues.

noggin/controller/user.py
```diff
@@ -188,3 +189,49 @@ def user_settings_otp_add(ipa, username):
...
+                'otptoken_find', [], {'ipatokenowner': username, 'all': True}
+            )['result']
+        ]
+        enabled_count = 0
+        for otp in tokens:
+            if not otp.disabled:
+                enabled_count += 1
+                if enabled_count > 1: break
+
+        if enabled_count > 1:
```
ryanlerch: Do you think we could move this logic to the otp list page, then not show the form if there is only 1 enabled otp key?

Pretty sure IPA won't let us disable the last token, even if the user tries to submit an outdated form.

ryanlerch (approved):

PR #163, opened 2020-03-16
Add the ability to Delete an otp token

### LINKED ISSUE #149: As a User, I want to be able to delete an OTP token, so that I can revoke lost or inactive tokens
**Acceptance Criteria**
- the page should be authenticated
- it should not be possible to delete the last token (as FreeIPA will [refuse the request](https://www.freeipa.org/page/V4/OTP#User-Managed_Tokens))
- the page should mention that if they want to go back to password auth only, then they should open an infra ticket (only when there's just one token left).

**Definition of Done**
- All of the above works
- Documentation is written
- CI passes

ryanlerch:
Two separate commits:

1. Remove hiding of disable button: Previously, it was thought that we could not disable or delete the last OTP token. We were mistaken, and this removes the logic that hides the disable button when only one active OTP tokenis present
2. add the ability to delete OTP tokens: Update the otp tokens list with the ability to delete tokens. (Resolves: #149)

### DIFF
--- a/noggin/controller/user.py
+++ b/noggin/controller/user.py
@@ -7,6 +7,7 @@ from noggin.form.edit_user import (
     UserSettingsKeysForm,
     UserSettingsAddOTPForm,
     UserSettingsDisableOTPForm,
+    UserSettingsDeleteOTPForm,
 )
 from noggin.representation.group import Group
 from noggin.representation.user import User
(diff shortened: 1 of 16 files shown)


noggin/controller/user.py
```diff
@@ -139,12 +140,6 @@ def user_settings_otp(ipa, username):
             'otptoken_find', [], {'ipatokenowner': username, 'all': True}
         )['result']
     ]
-    enabled_count = 0
```
StephenCoady, line 142: Maybe I misunderstood the conversation but won't freeipa reject the disabling of the last token when the auth policy is 2fa? We should probably revert to the flash message instead to allow the cases where they can disable the last token.

ryanlerch: Ok, have tested two possibilities here. 

1. on the freeIPA setup, i checked the Default Authentication types to be both "Password" and "Two factor authentication (password + OTP)". this allowed the following (note that this is just for logging into freeIPA webui / noggin, not kinit, which i didnt test)
 * without an otp token setup, i log in with just password.
 * with an otp token setup, i cannot log in with just password, i have to use password + OTP.
 * i can disable / delete the last OTP token, and then just log in with password again

2. on the freeIPA setup, i checked the Default Authentication types to be just"Two factor authentication (password + OTP)". this allowed the following (note that this is just for logging into freeIPA webui / noggin, not kinit, which i didnt test)
* before adding an otp token, i can log in with just password.
* when i add an otp token, i then have to log in with password  + OTP
* i *cannot* disable or delete the last OTP token. so i can never go back to just using a password.

ryanlerch: Assuming that we are setting up as option 2 above, as-is, if you try to disable or delete the last token, you get a flash message saying that it can't be done. Will work on some better flash messaging here, catching the error where we can't delete / disable the last one.

ryanlerch:
Okies, have updated the PR, so now if the mode is set to *just* Password+OTP, and you try to delete or disable the last token, you get a flash error (originated from IPA) saying you cant do that.

StephenCoady (approved):
looks good to me. tested the various scenarios with/without the correct policies in place and it all behaves as expected.

PR #167, opened 2020-03-18
added ability to sync tokens

### LINKED ISSUE #150: As a User, I want to be able to resync an OTP token, so that I can keep using it
**Acceptance Criteria**
- the page should **not** be authenticated
- it should use the `otptoken_sync` API call

**Definition of Done**
- All of the above works
- Documentation is written
- CI passes

StephenCoady:

resolves: #150


noggin/controller/authentication.py
```diff
@@ -53,3 +56,48 @@ def login():
...
+            'first_code': form.first_code.data,
+            'second_code': form.second_code.data,
+            'token': form.token.data,
+        }
+
+        url = (
+            'https://'
+            + random.choice(app.config['FREEIPA_SERVERS'])
+            + '/ipa/session/sync_token'
+        )
```
ryanlerch: do we need to reimplement this code here? can me not use `ipa_admin` or (most likely)  `untouched_ipa_client` here instead, like we do on the register page?

This might also remove the need for using requests here, no?

noggin/templates/_login_form.html
```diff
@@ -8,9 +8,9 @@
     <div class="form-group mb-0">
       {{ macros.with_errors(login_form.password, class="validate", placeholder="Password", tabindex="2", label=False) }}
       {% if lost_otp_token is not defined %}
-      <small><a href="{{ url_for('forgot_password_ask') }}">Forgot password?</a></small>
+      <small><a href="{{ url_for('forgot_password_ask') }}">Forgot password?</a></small><br><small><a href="{{ url_for('otp_sync') }}">Sync Token</a></small>
       {% else %}
-      <small><a href="{{ url_for('forgot_password_ask') }}">Forgotten password or lost OTP token?</a></small>
+      <small><a href="{{ url_for('forgot_password_ask') }}">Forgotten password or lost OTP token?</a></small><br><small><a href="{{ url_for('otp_sync') }}">Sync Token</a></small>
```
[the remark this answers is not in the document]
abompard: Also, do we want the two links in a vertical list, with a `br` tag in between? @ryanlerch , any opinion here? I heard `br` tags were considered uncool these days... ;-)

abompard (changes requested):

noggin/security/ipa.py
```diff
@@ -97,6 +98,35 @@ def otptoken_del(self, ipatokenuniqueid):
...
+            'password': password,
+            'first_code': first_code,
+            'second_code': second_code,
+            'token': token,
+        }
+        try:
+            url = "https://" + self._host + "/ipa/session/sync_token"
+            response = requests.post(url=url, data=data, verify=self._verify_ssl)
+        except requests.RequestException:
+            raise
```
abompard: I don't understand what this is for... If there is a `RequestException` you'll probably want to convert it to something else, no? Right now it does not do anything, the try..except is useless.

StephenCoady: So I wanted to just raise the exception so that the controller can handle it as I still have that logic there. Mainly because the controller has access to `app` so I can log from there. Is this a bad pattern? If I don't try...except here then it would make its way back to the controller anyway is it?

abompard: Yes, uncaught exceptions will bubble up automatically. This pattern here is basically useless.
My thinking is that you may want to convert the RequestException into an exception that is more specific to `python_ipa`, probably `python_ipa.exceptions.BadRequest` or a more appropriate subclass of it. Then you would catch that exception in the controller.

noggin/security/ipa.py
```diff
@@ -97,6 +98,35 @@ def otptoken_del(self, ipatokenuniqueid):
...
+        data = {
+            'user': user,
+            'password': password,
+            'first_code': first_code,
+            'second_code': second_code,
+            'token': token,
+        }
+        try:
+            url = "https://" + self._host + "/ipa/session/sync_token"
+            response = requests.post(url=url, data=data, verify=self._verify_ssl)
```
abompard: Any reason why you're not using `self._session` here instead of `requests` directly?

abompard (changes requested):

noggin/security/ipa.py
StephenCoady: Makes sense. I've updated this file but I haven't transformed the error into a `python_freeipa` error as its not directly related to the client and its still a requests error. I don't think they will be moving from a http request for syncing.

noggin/security/ipa.py
StephenCoady: Changed this now.

noggin/controller/authentication.py
```diff
@@ -57,3 +59,40 @@ def login():
...
+                    'Something went wrong trying to sync OTP token.', category='danger'
+                )
+                return redirect(url_for('root'))
+
+        except requests.RequestException as e:
+            app.logger.error(
+                f'A HTTP post error {e.__class__.__name__} happened while syncing a token for user '
+                f'{form.username}: {e}'
+            )
+            flash('Something went wrong trying to sync OTP token.', category='danger')
```
abompard: Same comment here about filling up the form's errors instead of using `flash` and redirecting.

noggin/controller/authentication.py
```diff
@@ -57,3 +59,40 @@ def login():
...
+                first_code=form.first_code.data,
+                second_code=form.second_code.data,
+                token=form.token.data,
+            )
+
+            if (
+                response.status_code == 200
+                and "Token sync rejected" not in response.text
+            ):
+                flash('Token successfully synchronized', category='success')
```
abompard, line 58: There should be a redirect here instead of at the bottom, if we start using the form's errors.

abompard (changes requested):

noggin/controller/authentication.py
```diff
@@ -57,3 +59,40 @@ def login():
...
+            else:
+                flash(
+                    'Something went wrong trying to sync OTP token.', category='danger'
+                )
```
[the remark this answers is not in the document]
StephenCoady: Unfortunately very little from the backend. Even the errors are dressed as 200 http status codes. The error only says that something was incorrect. I've updated the non field errors to reflect that now as you suggested.

noggin/controller/authentication.py
```diff
@@ -57,3 +59,40 @@ def login():
...
+                return redirect(url_for('root'))
+
+        except requests.RequestException as e:
```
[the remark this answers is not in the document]
StephenCoady: I've changed this now to use `python_freeipa`'s `BadRequest` as you suggested.

noggin/controller/authentication.py
```diff
@@ -65,34 +64,31 @@ def login():
...
-
-            if (
-                response.status_code == 200
-                and "Token sync rejected" not in response.text
-            ):
-                flash('Token successfully synchronized', category='success')
-            else:
-                flash(
-                    'Something went wrong trying to sync OTP token.', category='danger'
+        with handle_form_errors(form):
```
abompard, line 47: Yeah I guess you can use `handle_form_errors` here, it's not entirely necessary because you only have one set of exception handlers, but it'll work.
If the reason for the existence of `handle_form_errors` is not clear, please say so and I'll improve the documentation (which is severely lacking).

noggin/controller/authentication.py
```diff
@@ -65,34 +64,31 @@ def login():
...
-                return redirect(url_for('root'))
 
-        except requests.RequestException as e:
-            app.logger.error(
-                f'A HTTP post error {e.__class__.__name__} happened while syncing a token for user '
-                f'{form.username}: {e}'
-            )
-            flash('Something went wrong trying to sync OTP token.', category='danger')
-            return redirect(url_for('root'))
+                if response.ok and "Token sync rejected" not in response.text:
```
abompard: Actually this test could be done in the IPA client class method. Check that the request is OK and that the response text does not contain the "Token sync rejected" string in there, and raise a BadRequest error if it does. This way the controller only has to catch the BadRequest error.

noggin/tests/unit/utilities.py
```diff
@@ -30,4 +30,5 @@ def assert_form_generic_error(response, expected_message):
     page = BeautifulSoup(response.data, 'html.parser')
     error_message = page.select_one("#formerrors .text-danger")
     assert error_message is not None
+    print(error_message)
```
abompard: Looks like a leftover.

abompard (changes requested):

noggin/controller/authentication.py
StephenCoady: maybe I do misunderstand then, because when I don't use it raising the `FormError` actually raises an error and the application shows an exception as its not caught anywhere now.

StephenCoady:
@abompard I've pushed your suggestions and left the `handle_form_errors` line in because it saves me handling the exception. if you think I should take it out let me know. Other than that hopefully this is ready now.

noggin/tests/unit/conftest.py
```diff
@@ -113,3 +113,17 @@ def logged_in_dummy_user(client, dummy_user):
     ipa.logout()
     with client.session_transaction() as sess:
         sess.clear()
+
+
+@pytest.fixture
+def dummy_user_with_2fa(client, dummy_user):
```
abompard: I've merged a change to master that has a fixture pretty similar to this one (called `dummy_user_with_otp`), could you please rebase and use it?

noggin/form/sync_token.py
```diff
@@ -0,0 +1,24 @@
...
+
+    password = PasswordField(
+        'Password', validators=[DataRequired(message='You must provide a password')]
+    )
+
+    first_code = PasswordField(
+        'First OTP', validators=[DataRequired(message='You must provide a first code')]
+    )
+
+    second_code = PasswordField(
```
abompard: I wonder if it's useful to have a `PasswordField` here, since the tokens are one-time only. On most websites with OTP support that I've seen, I think the token field was clear text. What do you think?

abompard (commented):
Almost there, just one more question!

noggin/form/sync_token.py
StephenCoady: sure, that makes sense. pushed now!

abompard (changes requested):
Now that there's a controller in `controller/authentication.py` again, could you add it back to `noggin/app.py` please?

StephenCoady:
@abompard would it be better to move it somewhere else if you wanted to keep `authentication.py` controller-less?

abompard:
I thought of that but didn't find a better place. I'm open to suggestions though :-)

StephenCoady:
@abompard I couldn't find anywhere better than auth, and it is still related to there I suppose. Ready for review again.

abompard (approved):
LGTM! Thanks a lot for your work :-)

PR #168, opened 2020-03-18
added box for lost token to fas template

### LINKED ISSUE #151: As a User, I want to have a process to recover my account if I have enabled 2FA and lost my token
There should be a page (or a box on a page) that explains what to do when a user has lost their token. Currently, if a GPG key has been declared, they must send a signed email to Fedora admins with their request.

I propose that we mandate having a GPG key setup to enable 2FA. But we still need to think about the cases where a user with 2FA would have no GPG key (because it was expired or revoked maybe).

@relrod , @nirik , @puiterwijk  do you have an opinion?

StephenCoady:
Opening this PR as it seems to be the decided-upon solution in the linked issue. Also for feedback on where I have placed the box explaining what needs to be done.

resolves: #151

### DIFF
--- a/noggin/themes/fas/templates/main.html
+++ b/noggin/themes/fas/templates/main.html
@@ -103,3 +103,19 @@
         </a>
     </div>
 {% endmacro %}
+
+{% macro lost_otp_token() %}
+<div class="container py-4">
+    <div class="row">
+      <div class="col-6 mx-auto">
+        <div class="card">
+            <div class="card-body">
+              <h5 id="pageheading">Did you lose your OTP token?</h5>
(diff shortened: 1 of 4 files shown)


noggin/themes/fas/templates/main.html
```diff
@@ -103,3 +103,19 @@
...
 {% endmacro %}
+
+{% macro lost_otp_token() %}
+<div class="container py-4">
+    <div class="row">
+      <div class="col-6 mx-auto">
+        <div class="card">
+            <div class="card-body">
+              <h5 id="pageheading">Did you lose your OTP token?</h5>
+              <p>If you have lost your OTP token you need to send an email to <a href="mailto:admin@fp.o?Subject=Lost%20OTP" target="_top">admin@fp.o</a>.
```
abompard: This should be the actual email address, `fp.o` → `fedoraproject.org`

abompard (changes requested):

ryanlerch (approved):

PR #169, opened 2020-03-19
Require a password when creating a new OTP token

### LINKED ISSUE #164: As a User, I want to have to reauthenticate when enrolling an OTP token
As a User, I want to have to reauthenticate when enrolling a new OTP token, to ensure that only the user can add new tokens.

If a token is already added, the authentication method would be password + token

ryanlerch:
Resolves: #164 


abompard (approved):
There are a few weird things that are unrelated to the code you changed, so I think I'll fix them in a later PR

PR #172, opened 2020-03-21
localization support requires application internationalization

Jibec:
i18n contains two features:
. to be able to extract strings to translate
. to be able to usetranslated strings

Here is a demo of first step.
discussion is to be held in #173


ryanlerch:
I'm happy to merge this one (once the conflicts are resolved and tests pass)

For there, we can work on adding all the other strings, but this shows well how to add strings to both the code and templates.

abompard:
Included in #190

Jibec:
Thanks a lot for your work <3

PR #174, opened 2020-03-24
Remove requirement to have a GPG key for OTP & make OTP disable tests better

### LINKED ISSUE #155: As a User who has enrolled a 2FA token, I want to be unable to remove the last GPG key ID, so that I can reset my token with a signed email if necessary
**Acceptance Criteria**
- the user modification page refuses to remove the last GPG key ID if the user has a 2FA token enrolled.

**Definition of Done**
- All of the above works
- Documentation is written
- CI passes

ryanlerch:
Previously, we implemented the rule that a user needed
a GPG key to create an OTP token. It was decided that
this was not needed, so this commit removes that.

Closes: #155

Also added another commit that makes the otp tests realer:
```
The test_user_settings_otp_disable and
test_user_settings_otp_delete tests add two otp tokens, for
the purposes of removing (or disabling) one of them. However,
after adding the first token, to add the second token, the OTP would
have had to be supplied with the password. This was previously worked
around by the VCR cassette being generated on a system without OTP
enabled.

This PR updates the cassettes and checks that the OTP actaully works.
```


abompard (approved):
Thanks!

PR #175, opened 2020-03-26
remove redirect from user settings edit

ryanlerch:
previously, when editng the user settings (profile or gpg keys)
after a successful POST, noggin would redirect back to the profile
page, rather than to the settings page like with other settings pages
like OTP. This changes noggin to return to the settings edit form
rather than the profile page

### DIFF
--- a/noggin/controller/user.py
+++ b/noggin/controller/user.py
@@ -1,6 +1,6 @@
 from urllib.parse import urlparse
 
-from flask import flash, redirect, render_template, session, url_for
+from flask import flash, redirect, render_template, session, url_for, Markup
 import python_freeipa
 
 from noggin import app
@@ -41,7 +41,7 @@ def user(ipa, username):
     )
 
 
-def _user_mod(ipa, form, username, details):
+def _user_mod(ipa, form, username, details, redirect_to):
     with handle_form_errors(form):
         try:
             ipa.user_mod(username, **details)
@@ -53,8 +53,14 @@ def _user_mod(ipa, form, username, details):
                     f'An error happened while editing user {username}: {e.message}'
                 )
                 raise FormError("non_field_errors", e.message)
-        flash('Profile has been succesfully updated.', 'success')
-        return redirect(url_for('user', username=username))
+        flash(
+            Markup(
+                f'Profile Updated: <a href=\"{url_for("user", username=username)}\">'
+                'view your profile</a>'
+            ),
+            'success',
+        )
+        return redirect(url_for(redirect_to, username=username))
 
 
 @app.route('/user/<username>/settings/profile/', methods=['GET', 'POST'])
@@ -82,6 +88,7 @@ def user_settings_profile(ipa, username):
                 'fasgitlabusername': form.gitlab.data.lstrip('@'),
                 'fasrhbzemail': form.rhbz_mail.data,
             },
+            "user_settings_profile",
         )
         if result:
             return result
@@ -104,6 +111,7 @@ def user_settings_keys(ipa, username):
             form,
             username,
             {'ipasshpubkey': form.sshpubkeys.data, 'fasgpgkeyid': form.gpgkeys.data},
+            "user_settings_keys",
         )
         if result:
             return result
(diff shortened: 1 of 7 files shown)


abompard:
I think that on POST success it should still do a redirect, even if it is to the same page. Otherwise when trying to reload the page users will have the browser's warning message about sending the request again.
In my understanding the only situation where POST should not end up in a redirect is when there is an error with the posted data (in which case it should ideally never end up in a redirect but sometimes that makes things very complicated so we do it sometimes).

ryanlerch:
Thanks @abompard have updated with changes

noggin/controller/user.py
```diff
@@ -80,9 +86,8 @@ def user_settings_profile(ipa, username):
                 'fasgitlabusername': form.gitlab.data.lstrip('@'),
                 'fasrhbzemail': form.rhbz_mail.data,
             },
+            "user_settings_profile",
         )
-        if result:
-            return result
```
abompard, line 87: If you're not returning the redirect that `_user_mod` returned, then is it still redirecting?

noggin/controller/user.py
```diff
@@ -97,14 +102,13 @@ def user_settings_keys(ipa, username):
...
-        result = _user_mod(
+        _user_mod(
             ipa,
             form,
             username,
             {'ipasshpubkey': form.sshpubkeys.data, 'fasgpgkeyid': form.gpgkeys.data},
+            'user_settings_keys',
         )
-        if result:
-            return result
```
abompard, line 109: Same question here.

noggin/tests/unit/controller/test_user.py
```diff
@@ -75,11 +75,12 @@ def test_user_edit(client, logged_in_dummy_user):
...
     """Test posting to the user edit page: /user/<username>/settings/profile/"""
     result = client.post('/user/dummy/settings/profile/', data=POST_CONTENTS)
-    assert_redirects_with_flash(
+    assert_flashed_message(
         result,
         expected_url="/user/dummy/",
-        expected_message="Profile has been succesfully updated.",
+        expected_message="Profile Updated: <a href=\"/user/dummy/\">view your profile</a>",
         expected_category="success",
+        redirect=False,
```
abompard: If the POST is successful, then it should redirect, no?

noggin/tests/unit/controller/test_user.py
```diff
@@ -106,7 +107,7 @@ def test_user_edit_post_no_change(client, logged_in_dummy_user):
     # Do it once
     result = client.post('/user/dummy/settings/profile/', data=POST_CONTENTS)
 
-    assert result.status_code == 302
+    assert result.status_code == 200
```
abompard: Same thing here, if the POST is successful, it should redirect.

noggin/tests/unit/controller/test_user.py
```diff
@@ -142,11 +143,12 @@ def test_user_settings_keys(client, logged_in_dummy_user):
...
     """Test posting to the user edit page: /user/<username>/settings/keys/"""
     result = client.post('/user/dummy/settings/keys/', data=POST_CONTENTS_KEYS)
-    assert_redirects_with_flash(
+    assert_flashed_message(
         result,
         expected_url="/user/dummy/",
-        expected_message="Profile has been succesfully updated.",
+        expected_message="Profile Updated: <a href=\"/user/dummy/\">view your profile</a>",
         expected_category="success",
+        redirect=False,
```
abompard: Same here.

noggin/tests/unit/controller/test_user.py
```diff
@@ -173,7 +175,7 @@ def test_user_settings_keys_post_no_change(client, logged_in_dummy_user):
     # Do it once
     result = client.post('/user/dummy/settings/keys/', data=POST_CONTENTS_KEYS)
 
-    assert result.status_code == 302
+    assert result.status_code == 200
```
abompard: And same here. Unless I'm missing something?

noggin/tests/unit/utilities.py
```diff
@@ -5,17 +5,22 @@
 import pyotp
 
 
-def assert_redirects_with_flash(
-    response, expected_url, expected_message, expected_category
+def assert_flashed_message(
+    response, expected_url, expected_message, expected_category, redirect=True
```
abompard: I haven't seen a case where there should be no redirect.

abompard (changes requested):

abompard (approved):
LGTM, thanks!

PR #190, opened 2020-04-02
Add translation support

ryanlerch:
This takes the commit from PR #172 and updates it to make translation fully functional. #172 was a proof of concept of extracting the strings, but we need to fully functioning to be able to pass the tests, so here we are.

I have not flagged many more strings for translation, figured that I could do that in my next PR. 

a couple of notes:

* the comment in the commit of #172 about f-strings still applies. it looks like we will have to convert these back to format()

* the translated strings in this PR were done with an automated translation tool. Apologies if these translations make you laugh.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,6 @@ docs/_build/
 
 # Vagrant
 .vagrant
+
+# Translations
+*.mo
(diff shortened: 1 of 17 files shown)


abompard:
Do we want to ship the binary `.mo` files?

ryanlerch:
> Do we want to ship the binary `.mo` files?

okies, have removed them, and added a command in the vagrant setup to build them. and added *.mo to the gitignore file.

in development, they can also just be rebuilt with:

    poetry run pybabel compile -d noggin/translations

abompard (approved):
Cool, let's translate more strings now :-)

PR #191, opened 2020-04-02
added the ability to hide certain groups

### LINKED ISSUE #189: Disallow leaving 'ipausers' or other critical groups
Currently I can leave the default 'ipausers' group, is that expected?

There might be other automated groups we want to not allow people to leave too.

StephenCoady:

resolves #189

### DIFF
--- a/noggin/controller/root.py
+++ b/noggin/controller/root.py
@@ -3,7 +3,6 @@ from flask import render_template, request, redirect, url_for, session, jsonify
 from noggin import app
 from noggin.form.register_user import RegisterUserForm
 from noggin.form.login_user import LoginUserForm
-from noggin.representation.group import Group
 from noggin.representation.user import User
 from noggin.security.ipa import maybe_ipa_session
 from noggin.utility import with_ipa, handle_form_errors
@@ -69,7 +68,7 @@ def search_json(ipa):
             res.append({'uid': user_.username, 'cn': user_.name})
 
     if groupname:
-        groups_ = [Group(g) for g in ipa.group_find(groupname)['result']]
+        groups_ = ipa.group_find(groupname)
         for group_ in groups_:
             res.append({'cn': group_.name, 'description': group_.description})
 
--- a/noggin/controller/root.py
+++ b/noggin/controller/root.py
@@ -4,6 +4,8 @@ from noggin import app
 from noggin.form.register_user import RegisterUserForm
 from noggin.form.login_user import LoginUserForm
 from noggin.representation.user import User
+from noggin.representation.group import Group
+
 from noggin.security.ipa import maybe_ipa_session
 from noggin.utility import with_ipa, handle_form_errors
 
@@ -68,7 +70,7 @@ def search_json(ipa):
             res.append({'uid': user_.username, 'cn': user_.name})
 
     if groupname:
-        groups_ = ipa.group_find(groupname)
+        groups_ = [Group(g) for g in ipa.group_find(groupname)['result']]
         for group_ in groups_:
             res.append({'cn': group_.name, 'description': group_.description})
 
--- a/noggin/controller/root.py
+++ b/noggin/controller/root.py
@@ -3,8 +3,8 @@ from flask import render_template, request, redirect, url_for, session, jsonify
 from noggin import app
 from noggin.form.register_user import RegisterUserForm
 from noggin.form.login_user import LoginUserForm
-from noggin.representation.user import User
 from noggin.representation.group import Group
+from noggin.representation.user import User
 
 from noggin.security.ipa import maybe_ipa_session
 from noggin.utility import with_ipa, handle_form_errors
--- a/noggin/controller/user.py
+++ b/noggin/controller/user.py
@@ -12,7 +12,6 @@ from noggin.form.edit_user import (
     UserSettingsDisableOTPForm,
     UserSettingsDeleteOTPForm,
 )
-from noggin.representation.group import Group
 from noggin.representation.user import User
 from noggin.representation.otptoken import OTPToken
 from noggin.security.ipa import maybe_ipa_login
@@ -32,16 +31,8 @@ def user(ipa, username):
     # As a speed optimization, we make two separate calls.
     # Just doing a group_find (with all=True) is super slow here, with a lot of
     # groups.
-    hidden_groups = app.config.get('HIDDEN_GROUPS')
-    groups = [
-        Group(g)
-        for g in ipa.group_find(user=username, all=False)['result']
-        if Group(g).name not in hidden_groups
-    ]
-    managed_groups = [
-        Group(g)
-        for g in ipa.group_find(membermanager_user=username, all=False)['result']
-    ]
+    groups = ipa.group_find(user=username, all=False)
+    managed_groups = ipa.group_find(membermanager_user=username, all=False)
     return render_template(
         'user.html', user=user, groups=groups, managed_groups=managed_groups
     )
--- a/noggin/controller/user.py
+++ b/noggin/controller/user.py
@@ -32,7 +32,12 @@ def user(ipa, username):
     # As a speed optimization, we make two separate calls.
     # Just doing a group_find (with all=True) is super slow here, with a lot of
     # groups.
-    groups = [Group(g) for g in ipa.group_find(user=username, all=False)['result']]
+    hidden_groups = app.config.get('HIDDEN_GROUPS')
+    groups = [
+        Group(g)
+        for g in ipa.group_find(user=username, all=False)['result']
+        if Group(g).name not in hidden_groups
+    ]
     managed_groups = [
         Group(g)
         for g in ipa.group_find(membermanager_user=username, all=False)['result']
--- a/noggin/controller/user.py
+++ b/noggin/controller/user.py
@@ -12,6 +12,7 @@ from noggin.form.edit_user import (
     UserSettingsDisableOTPForm,
     UserSettingsDeleteOTPForm,
 )
+from noggin.representation.group import Group
 from noggin.representation.user import User
 from noggin.representation.otptoken import OTPToken
 from noggin.security.ipa import maybe_ipa_login
@@ -31,8 +32,11 @@ def user(ipa, username):
     # As a speed optimization, we make two separate calls.
     # Just doing a group_find (with all=True) is super slow here, with a lot of
     # groups.
-    groups = ipa.group_find(user=username, all=False)
-    managed_groups = ipa.group_find(membermanager_user=username, all=False)
+    groups = [Group(g) for g in ipa.group_find(user=username, all=False)['result']]
+    managed_groups = [
+        Group(g)
+        for g in ipa.group_find(membermanager_user=username, all=False)['result']
+    ]
     return render_template(
         'user.html', user=user, groups=groups, managed_groups=managed_groups
     )
--- a/noggin/defaults.cfg
+++ b/noggin/defaults.cfg
@@ -7,4 +7,4 @@ USER_DEFAULTS = {"user_locale": "en-US", "user_timezone": "UTC"}
 THEME = "default"
 PASSWORD_RESET_EXPIRATION = 10  # in minutes
 PASSWORD_RESET_LOCK_DIR = "/tmp/noggin-pw-reset"
-HIDDEN_GROUPS = "hidden_groups"
+HIDE_GROUPS_IN = "hidden_groups"
--- a/noggin/defaults.cfg
+++ b/noggin/defaults.cfg
@@ -7,3 +7,4 @@ USER_DEFAULTS = {"user_locale": "en-US", "user_timezone": "UTC"}
 THEME = "default"
 PASSWORD_RESET_EXPIRATION = 10  # in minutes
 PASSWORD_RESET_LOCK_DIR = "/tmp/noggin-pw-reset"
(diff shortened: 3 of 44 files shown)


noggin/defaults.cfg
```diff
@@ -7,3 +7,5 @@ USER_DEFAULTS = {"user_locale": "en-US", "user_timezone": "UTC"}
 THEME = "default"
 PASSWORD_RESET_EXPIRATION = 10  # in minutes
 PASSWORD_RESET_LOCK_DIR = "/tmp/noggin-pw-reset"
+# groups to hide from the user. should in the format "group1","group2" 
+HIDDEN_GROUPS = "ipausers"
```
abompard: This should be a list

noggin/controller/user.py
```diff
@@ -36,8 +36,13 @@ def user(ipa, username):
         Group(g)
         for g in ipa.group_find(membermanager_user=username, all=False)['result']
     ]
+    hidden_groups = app.config.get('HIDDEN_GROUPS')
+    complete_group_list = [group for group in groups if group.name not in hidden_groups]
```
abompard: You could do this filtering directly in the list comprehension that sets `groups` above.

abompard (changes requested):
A couple comments, also please add a unit test.

noggin/tests/unit/conftest.py
```diff
@@ -40,6 +40,7 @@ def client(ipa_cert):
     app.config['TESTING'] = True
     app.config['DEBUG'] = True
     app.config['WTF_CSRF_ENABLED'] = False
+    app.config['HIDDEN_GROUPS'] = ['ipausers']
```
abompard: Are you sure you need to override it here? I thought the defaults were loaded for the testing mode too, no?

noggin/tests/unit/controller/test_user.py
```diff
@@ -186,3 +186,19 @@ def test_user_settings_keys_post_bad_request(client, logged_in_dummy_user):
         )
         result = client.post('/user/dummy/settings/keys/', data=POST_CONTENTS_KEYS)
     assert_form_generic_error(result, 'something went wrong')
+
+
+def test_user_cant_see_hidden_groups(client, logged_in_dummy_user):
```
abompard: You're missing the VCR decorator (and the cassette that it will generate)

noggin/tests/unit/controller/test_user.py
```diff
@@ -186,3 +186,19 @@ def test_user_settings_keys_post_bad_request(client, logged_in_dummy_user):
...
+    result = client.get('/user/dummy/')
+    page = BeautifulSoup(result.data, 'html.parser')
+    assert page.title
+    assert page.title.string == 'dummy\'s Profile - noggin'
+    assert page.select_one('.nav-link .badge-pill').get_text(strip=True) == '0'
+
+
+def test_user_can_see_dummy_group(client, dummy_user_as_group_manager):
```
abompard: Same here, you need the VCR decorator.

abompard (changes requested):

noggin/tests/unit/conftest.py
StephenCoady: I thought the same, but saw a failure without it. However now I can't reproduce, so maybe it was unrelated. I've removed it now.

abompard (changes requested):
We want those groups to be hidden everywhere in noggin, not just on the page that display a user's group. If you grep for `group_find` in the code, you'll find other places to filter.

noggin/security/ipa.py
```diff
@@ -35,6 +36,20 @@ class Client(IPAClient):
...
+    def group_find(self, *args, **kwargs):
+        """
+        Find a group. A wrapped method to allow filtering of groups to be hidden from the user.
+
+        TODO: this needs to be retained when we migrate to the newer Client class.
+        """
+        hidden_groups = current_app.config.get('HIDDEN_GROUPS')
+        groups = [
+            g
+            for g in IPAClient.group_find(self, *args, **kwargs)['result']
```
StephenCoady: @abompard I think this is ready now if you get a chance. I still need to use the Class name directly for some reason. If its a big deal I can find out why I can't use super.

abompard (changes requested):
So, I've thought of a different way of doing it. When I look at the shape of the dict returned by the `group_find` command, it also contains a count of the results and a message, which would ideally need to update to return the same dict shape. Then it occured to me, could we do the filtering in the request rather than in python after the request? Fortunately, IPA can have groups of groups, and the `group_find` command has an argument that allows excluding the groups that are in a certain group.

So what we can do is have a configuration variable for a specific group that we would put all the groups we want to hide into, and then use the appropriate argument. Basically our `group_find` method can be just:
```python
       kwargs.setdefault("not_in_groups", []).append(
           current_app.config.get('HIDE_GROUPS_IN')
       )
       return super().group_find(*args, **kwargs)
```
Then we need to go in the IPA server, create the group, and add as members all the groups we want to hide.

We could choose to exclude groups that are members of a specified list of groups but I think one is enough for what we want to do.

Second thing: we want Noggin users to have nothing to do with the hidden groups, so please also make sure they can't interact with them in `group_add_member` and `group_remove_member`. Otherwise it would for example still be possible to remove oneself from `ipausers` even if it's not listed in one's groups by sending the right POST request to the right endpoint. I think having unit tests for that is important.

noggin/controller/root.py
```diff
@@ -3,8 +3,9 @@
 from noggin import app
 from noggin.form.register_user import RegisterUserForm
 from noggin.form.login_user import LoginUserForm
-from noggin.representation.group import Group
 from noggin.representation.user import User
+from noggin.representation.group import Group
+
```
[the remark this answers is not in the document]
StephenCoady: Leftover from removing this representation from this file.

StephenCoady:
@abompard thanks for the new way to do it. I've implemented it that way now. It had a fairly far reaching impact on testing though because I needed to change the group_or_404 method. Hopefully you are OK with what I've done there. If not let me know.

noggin/security/ipa.py
```diff
@@ -35,6 +37,15 @@ class Client(IPAClient):
     TODO: send this upstream.
     """
 
+    def group_find(self, *args, **kwargs):
+        """
+        Find a group. A wrapped method to allow filtering of groups to be hidden from the user.
+
+        TODO: this needs to be retained when we migrate to the newer Client class.
+        """
+        kwargs.setdefault("not_in_group", current_app.config.get('HIDDEN_GROUPS'))
```
abompard: If you do it that way, a call to `group_find(not_in_group="blah")` will only be excluding groups that are in the `blah` group, not the one in the config. That's why I suggested something different in my last review. I know the `setdefault()` method is not very obvious to use, feel free to do it differenly if you prefer.

noggin/tests/unit/controller/test_group.py
```diff
@@ -30,6 +30,24 @@ def test_groups_list(client, logged_in_dummy_user, dummy_group):
...
 
+@pytest.mark.vcr()
+def test_groups_list_no_hidden(client, logged_in_dummy_user, dummy_group):
+    """Test the groups list: /groups/"""
+    result = client.get('/groups/')
+    assert result.status_code == 200
+    page = BeautifulSoup(result.data, 'html.parser')
+    groups = page.select("ul.list-group li")
+    group_names = [g.find("span", class_="title").get_text(strip=True) for g in groups]
+    assert "ipausers" not in group_names
```
abompard, line 43: Ideally our testing system would ensure that the hidden group exists on the server and put the `ipausers` group in there. A session-wide fixture could to the trick, so it's only done once per run.

noggin/utility/__init__.py
```diff
@@ -60,10 +60,11 @@ def fn(*args, **kwargs):
 
 
 def group_or_404(ipa, groupname):
-    try:
-        return ipa.group_show(groupname)
-    except python_freeipa.exceptions.NotFound:
+    group = ipa.group_find(groupname)['result']
+    if ipa.group_find(groupname)['result'] == []:
```
abompard: You're doing the call twice, why not just test `group` ? Also, you can test that a list is empty by just doing `if not groups:`

noggin/utility/__init__.py
```diff
@@ -60,10 +60,11 @@ def fn(*args, **kwargs):
...
         abort(404)
```
abompard: Maybe add a message, like `abort(404, f"Group {groupname} could not be found.")`

abompard (changes requested):
Thanks, it's starting to look good! I still have a few comments :-)

noggin/utility/__init__.py
StephenCoady: Ah, I added the var for that reason and just forgot to use it :man_facepalming:

noggin/security/ipa.py
StephenCoady: Yeah, it wasn't so much that. It was that `not_in_group` can only take one group. If we want to still allow a different exclusion somewhere in our app, that means I'll need to make the call twice and add those groups together. that seems a bit weird to me becuase they may not be mutually exclusive.

StephenCoady: Do you see us using `not_in_group` anywhere else?

noggin/tests/unit/controller/test_group.py
StephenCoady: OK, I've added that fixture and pushing now. Am I right in saying that this test won't change though and you just mean that our test can reliably check that the group exists? Or am I missing something?

noggin/security/ipa.py
abompard: Actually, `not_in_group` can take a list of groups, I just tested it. Did it not work for you? I don't however think we're ever going to use a list of groups here, so I guess a single value is OK.

However I think it would be nice to remove the "hidden-groups" group from the output, and for that I think we can set the `posix` keyword argument. I'll only output posix groups, and I think that's the only ones we need.

By the way, setting `posix` would also filter out the `ipausers` group, so it would be sufficient, but I still think it's useful to be able to filter out unwanted posix groups, so let's keep the `not_in_group` option.

noggin/tests/unit/controller/test_group.py
abompard: Yes, that's the idea.

noggin/tests/unit/conftest.py
```diff
@@ -35,6 +35,15 @@ def ipa_cert():
         yield
 
 
+@pytest.fixture(scope="session")
+def add_hidden_group():
+    hidden_group = app.config.get('HIDDEN_GROUPS')
+    parent_group = ipa_admin.group_find(hidden_group)
+    if not parent_group:
+        ipa_admin.group_add(cn=hidden_group)
```
abompard: Please also set the `nonposix` keyword argument to `True`, we don't need that group to be posix.

noggin/tests/unit/controller/test_group.py
```diff
@@ -30,6 +30,24 @@ def test_groups_list(client, logged_in_dummy_user, dummy_group):
     assert group_desc.get_text(strip=True) == "A dummy group"
 
 
+@pytest.mark.vcr()
+def test_groups_list_no_hidden(client, logged_in_dummy_user, dummy_group):
```
abompard: I think those tests still need to depend on the `add_hidden_group` fixture.

noggin/tests/unit/conftest.py
```diff
@@ -35,6 +35,20 @@ def ipa_cert():
         yield
 
 
+@pytest.fixture
+def add_hidden_group(logged_in_dummy_user):
+    hidden_group = app.config.get('HIDE_GROUPS_IN')
+    ipa = logged_in_dummy_user
+    # We need to use request here as our ipa.group_find method will filter the hidden groups out
```
abompard: There's another way you can do that. You can allow the `group_find` caller to override `posix=True`. You can do something like:
```python
final_kwargs = {"posix": True, "not_in_group": []}
final_kwargs.update(kwargs)
final_kwargs["not_in_group"].append(current_app.config.get('HIDE_GROUPS_IN'))
return super().group_find(*args, **kwargs)
```
Then if you pass `posix=False` it will override the defaults. Sorry to make you amend this function again! :-D

noggin/utility/__init__.py
```diff
@@ -60,10 +60,11 @@ def fn(*args, **kwargs):
 
 
 def group_or_404(ipa, groupname):
-    try:
-        return ipa.group_show(groupname)
-    except python_freeipa.exceptions.NotFound:
-        abort(404)
+    group = ipa.group_find(groupname)['result']
+    if not group:
+        abort(404, f"Group {groupname} could not be found.")
```
abompard: Please use i18n here now that we have it. :-)

abompard (changes requested):
Almost there! Sorry I have yet another comment!

99-fasschema.ldif
```diff
@@ -0,0 +1,32 @@
+#
```
abompard: It looks like this file was wrongly added with the last commit.

abompard (changes requested):

99-fasschema.ldif
StephenCoady: :man_facepalming:

abompard (approved):
LGTM! Thanks for your patience!

james02135 (approved):

PR #198, opened 2020-04-03
Add dirty form checking to user settings

### LINKED ISSUE #182: Change in profile settings is forgotten when tab is changed
## Description
When doing change in Profile settings and moving to another tab without saving changes the change is forgotten.

## Steps to reproduce
1. Login
2. Open Settings
3. Change anything in Profile
4. Switch to OTP tab

## Expected output
Dialog to ask if I want to save the changes.

## Current output
Change is silently forgotten.

ryanlerch:
This adds dirty form checking to user profile settings,
and user keys settings. If a user makes a change to these
forms, then tries to leave the page, a dialog is presented
alerting them to the fact they have unsaved changes.

Resolves: #182

### DIFF
--- a/noggin/templates/_form_macros.html
+++ b/noggin/templates/_form_macros.html
@@ -39,3 +39,8 @@
     </div>
   {% endif %}
 {% endmacro %}
+
+{% macro unsaved_changes() %}
+<script src="{{ url_for('static', filename='js/vendor/jquery.dirty-0.7.2/jquery.dirty.js') }}"></script>
+<script>$("form").dirty({preventLeaving: true})</script>
+{% endmacro %}
--- a/noggin/templates/user-settings-profile.html
+++ b/noggin/templates/user-settings-profile.html
@@ -24,3 +24,8 @@
 </form>
 {% endblock %}
 
+{% block scripts %}
+  {{ super () }}
+  {{ macros.unsaved_changes() }}
+{% endblock %}
+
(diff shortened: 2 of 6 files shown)


noggin/templates/_form_macros.html
```diff
@@ -39,3 +39,8 @@
     </div>
   {% endif %}
 {% endmacro %}
+
+{% macro unsaved_changes() %}
+<script src="{{ url_for('static', filename='js/vendor/jquery.dirty-0.7.2/jquery.dirty.js') }}"></script>
+<script>$("form").dirty({preventLeaving: true})</script>
```
abompard, line 45: You may want to wrap that in `$(...)` to make sure it's executed after the document has loaded, no?

noggin/templates/user-settings-profile.html
```diff
@@ -2,7 +2,7 @@
 
 {% import '_form_macros.html' as macros %}
 {% block settings_content %}
-<form class="needs-validation" action="{{ url_for('user_settings_profile', username=user.username) }}" method="post" novalidate>
+<form class="needs-validation" id="profile-form" action="{{ url_for('user_settings_profile', username=user.username) }}" method="post" novalidate>
```
abompard: Why set the id here?

noggin/templates/_form_macros.html
abompard: Also, I'm thinking we may want that on all our forms, don't we? If so we can do it at the top-level template.

noggin/templates/user-settings-profile.html
ryanlerch: ha! was originally selecting by that ID, then stopped. this was just a leftover...

noggin/templates/_form_macros.html
ryanlerch: There are some that we def dont want this on, like login, or register...

abompard (approved):
Thanks!

PR #199, opened 2020-04-03
make QR code modal show when document loads

### LINKED ISSUE #179: OTP QR Modal not showing instantly
When creating an OTP token, the modal that contains the QR code does not appear instantly, and the page is shown for sometimes a couple of seconds before the modal is shown.

ryanlerch:
Previoiusly, the method we were using for showing
the QR code otp modal was waiting for images and
other resources on the page to load before showing the
dialog. Now wer use the document.ready to show the modal
straight away

Resolves: #179

### DIFF
--- a/noggin/templates/user-settings-otp.html
+++ b/noggin/templates/user-settings-otp.html
@@ -79,7 +79,7 @@
   {% if otp_uri %}
   <script src="{{ url_for('static', filename='js/vendor/jquery-qrcode-1.0/jquery.qrcode.min.js') }}"></script>
   <script>
-    $(window).on('load',function(){
+    $(document).ready(function() {
         $('#otp-modal').modal('show');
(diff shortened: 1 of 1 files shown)


noggin/templates/user-settings-otp.html
```diff
@@ -79,7 +79,7 @@ <h5 id="pageheading">OTP Tokens</h5>
   {% if otp_uri %}
   <script src="{{ url_for('static', filename='js/vendor/jquery-qrcode-1.0/jquery.qrcode.min.js') }}"></script>
   <script>
-    $(window).on('load',function(){
+    $(document).ready(function() {
```
abompard, line 82: According to [the jQuery docs](https://api.jquery.com/ready/) you can also write `$(function() {...})`. Not sure which is the clearest, I was just mentioning it.

abompard (approved):

PR #201, opened 2020-04-06
Validate email addresses

### LINKED ISSUE #184: validate email before allowing account creation
Right now you can fill in the account creation information and then login right after. 

But this means: 

* You can use whatever email you want, you could use someones email you don't like and start behaving poorly and give them a bad reputation. 
* You could be a spammer and just create 10,000 accounts with bogus email addresses. 

We should validate their email before allowing the account creation to finish. Send an email with a activation link or token or something. 

Feel free to close if this is already filed or addressed. Thanks!

abompard:
This changeset modifies the registration workflow to validate email addresses. Here's an overview of the changes:

1. upon registration, we no longer create users but [stage users](https://www.freeipa.org/page/V4/User_Life-Cycle_Management#User_Story_2). Stage users can't login or have a password, so we don't ask for the password at that time.
2. an email with a link including a JWT token is sent to the provided email address.
3. on that link's landing page, the token is verified and the user is asked for a password
4. on form submission, the user is activated (moved from stage users to users) and the password is set as before.

This PR is not ready yet, I need to adjust the unit tests, but I wanted to post it here for workflow validation. What do you think? Did I miss something? Any attack vectors?

Resolves: #184


abompard:
OK, this should be ready for review now, there's only one failing test and it should be fixed by fedora-infra/freeipa-fas#89.

noggin/defaults.cfg
```diff
@@ -5,6 +5,9 @@ SESSION_COOKIE_HTTPONLY = True
 SESSION_COOKIE_SECURE = True
 USER_DEFAULTS = {"user_locale": "en-US", "user_timezone": "UTC"}
 THEME = "default"
+PASSWORD_POLICY = {
+    "min": 8, "max": -1
+}
```
ryanlerch, line 10: If the password policy is changed on the freeIPA side does it also have to be manually changed here on the Noggin side?

If so, i think we either need to get the password policy from IPA (if we can), or document this really well for system admins so we dont run into issues where the password policies dont match.

noggin/controller/registration.py
```diff
@@ -40,26 +66,26 @@ def _handle_registration_validation_error(username, e):
...
+    # First, create the stage user.
     try:
-        ipa_admin.user_add(
+        user = ipa_admin.stageuser_add(
             username,
             form.firstname.data,
             form.lastname.data,
-            f'{form.firstname.data} {form.lastname.data}',  # TODO ???
-            user_password=password,
+            # f'{form.firstname.data} {form.lastname.data}',  # TODO ???
```
ryanlerch: DO we need this commented out still (and the 'todo ???') or can it just be deleted?

ryanlerch (commented):
LGTM, just a couple of comments.

noggin/defaults.cfg
abompard: Yeah, I agree. There is a way to get the generic password policy from IPA or group-specific password policies, but I don't think Noggin has a way to determine which policy is going to be used for the users that it creates. Or maybe it does but I'm betting that would be pretty complex.
At first I wanted to let FreeIPA errors propagate back to the user but unfortunately FreeIPA only tells you that your password does not match the policy, it does not tell you what the policy is, so the user is left wondering what they have done wrong. I think it's less bad to ask ops to adjust the password policy in noggin when they change it in FreeIPA. Maybe there's a way to do it better but I think it'll do for now (I'll open a ticket for that when this is merged so we remember it).

noggin/controller/registration.py
abompard: Yeah I'll change it, I wanted to leave it there because there's an inconsistancy in the python_freeipa library: `user_add` requires the fullname, but `stageuser_add` does not. Anyway leaving a commented line is certainly not the way to remember this ;-)

PR #210, opened 2020-04-07
Translation strings

### LINKED ISSUE #173: As a User, I want to access FAS in my language
FAS is a mandatory portal for the Fedora community. It is the first tool you have to go through to join our community.

When english isn't your first language, understanding the password rules, how to join a group and to fill your user information is important to prevent discouragement.

Local community should also be able to use some of our applications for local activities, using loal language.

What should be done:

- internationalization support (extract strings to translate, provide localized content)
- localization support (allow translators to provide localized content) 
- automation (to push content to translators, to push translator work to users)

I did a demo of string extraction in this PR:
https://github.com/fedora-infra/noggin/pull/172

I can help for initialization, but after that, devs should directly include translation directives.
We should identify what is to translate and what isn't. In #172, I assumed app.logger.error should not be translated and that form.errors is to be translated.

This will impact the way to develop:

- Python's f-string aren't compatible with gettext
- Some things will be moved from templates to python files (variables are easier handled in python files than)
- Strings should not be fragmented (the user see a string, but the code is multiple prints)

This will impact the way to release:

- If automation isn't possible to push new translated content to FAS (like we have for Websites or Documentation), translators should be notified at least a week before release by email to translator mailing list.

Acceptance Criteria:

- As a user, FAS detect my language and display content in my language
- As a user, I can switch to another language manualy
- As a user, I can see language in URL
- As a translator, I don't need to ask devs to publish a new language (automation)

ryanlerch:
Flags all the strings in the python code, templates and theme templates for translation.

also removes an old directory "po/" that is not used, and updates the POT and french PO file with all the new strings.

resolves: #173


noggin/form/edit_user.py
```diff
@@ -1,4 +1,5 @@
 from flask_wtf import FlaskForm
+from flask_babel import lazy_gettext as _l
```
abompard: It's fine to just import it as `_`, no? We don't need the other one.

noggin/form/login_user.py
```diff
@@ -1,3 +1,4 @@
+from flask_babel import lazy_gettext as _l
```
abompard: Same comment

noggin/tests/unit/utility/test___init__.py
```diff
@@ -102,12 +102,12 @@ def test_formerror_unknown_field(client):
...
 
 def test_formerror_existing_field(client):
-    with current_app.test_request_context('/'):
+    with current_app.test_request_context('/', method="POST", data={"username": "1"}):
         form = LoginUserForm()
         form.validate()
-        error = FormError("username", "error message")
+        error = FormError("username", "You must provide a user name")
         error.populate_form(form)
-    assert form.errors["username"] == ["You must provide a user name", "error message"]
```
abompard, line 110: Actually, that's not what I was testing here (although documentation would have been nice). I'm testing that raising a `FormError` on a field that already has an error is appending to that field's errors list.

noggin/themes/fas/templates/main.html
```diff
@@ -1,5 +1,5 @@
 {% extends "master.html" %}
-{% block website %}Fedora Accounts{% endblock %}
+{% block website %}(Fedora Accounts){% endblock %}
```
abompard: Did you just want to add the parenthesis here?

abompard (changes requested):

abompard:
OK I've pushed a couple commits to fix what I've noticed because this is typically the kind of PR that will get more and more painful if we wait too long before merging it

abompard (approved):
OK let's merge this :-)

PR #213, opened 2020-04-08
Make OTP Token creation a little less cryptic

### LINKED ISSUE #181: It is not super obvious on how to register an OTP token
I was left a bit clueless when trying to add a new OTP token. I did not get that I needed to provide my password a second time, even tho I was already logged in.

I think adding a small "How To" section describing the steps needed would help.

ryanlerch:
This adds a little more information to the creation of OTP
tokens, providing a better experience for users.

This also makes the token description optional.

Resolves: #181


abompard (approved):
Looks much better, thanks. Just a minor line-too-long issue to fix.

PR #214, opened 2020-04-08
Allow the user to reenable disabled otp tokens

### LINKED ISSUE #194: As a user I want to be able to enable a token I have previously disabled so that I can use it again
We should allow users to re-enable an OTP token they have previously disabled. The `otptoken_mod` already exists in `ipa.py` so we will just need to add the button to the view and test it.

_Original question: Is there any way to activate a disabled token? If not, why keep them after disabling and how it will be different from deleting the token?_

ryanlerch:
Resolves: #194

### DIFF
--- a/noggin/controller/user.py
+++ b/noggin/controller/user.py
@@ -9,8 +9,7 @@ from noggin.form.edit_user import (
     UserSettingsProfileForm,
     UserSettingsKeysForm,
     UserSettingsAddOTPForm,
-    UserSettingsDisableOTPForm,
-    UserSettingsDeleteOTPForm,
+    UserSettingsOTPStatusChange,
 )
 from noggin.representation.group import Group
 from noggin.representation.user import User
@@ -188,7 +187,7 @@ def user_settings_otp(ipa, username):
 @with_ipa(app, session)
 @require_self
 def user_settings_otp_disable(ipa, username):
-    form = UserSettingsDisableOTPForm()
+    form = UserSettingsOTPStatusChange()
 
     if form.validate_on_submit():
         token = form.token.data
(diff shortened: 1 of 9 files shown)


noggin/controller/user.py
```diff
@@ -212,11 +211,36 @@ def user_settings_otp_disable(ipa, username):
...
+
+    if form.validate_on_submit():
+        token = form.token.data
+        try:
+            ipa.otptoken_mod(ipatokenuniqueid=token, ipatokendisabled=None)
+        except (
+            python_freeipa.exceptions.BadRequest,
+            python_freeipa.exceptions.FreeIPAError,
+        ) as e:
+            flash(f'Cannot enable the token. {e}', 'danger')
```
abompard: This should be i18n'ed ;-)

noggin/tests/unit/controller/test_user_otp.py
```diff
@@ -407,3 +407,93 @@ def test_user_settings_otp_delete_lasttoken(
...
+        "/user/dummy/settings/otp/disable/",
+        data={"token": tokenid},
+        follow_redirects=True,
+    )
+
+    page = BeautifulSoup(result.data, "html.parser")
+    tokenlist = page.select("div.list-group .list-group-item")
+
+    # check we are still showing 2 item
+    assert len(tokenlist) == 2
```
abompard, line 487: Maybe you also want to check here that the token is shown as disabled?

noggin/tests/unit/controller/test_user_otp.py
```diff
@@ -407,3 +407,93 @@ def test_user_settings_otp_delete_lasttoken(
...
+        "/user/dummy/settings/otp/enable/",
+        data={"token": tokenid},
+        follow_redirects=True,
+    )
+
+    page = BeautifulSoup(result.data, "html.parser")
+    tokenlist = page.select("div.list-group .list-group-item")
+
+    # check we are still showing 2 item
+    assert len(tokenlist) == 2
```
abompard, line 506: And here you may want to check that the token is shown as enabled?

abompard (changes requested):

abompard (approved):
LGTM, thanks!

PR #233, opened 2020-04-20
make loggedin change password form otp aware

### LINKED ISSUE #206: Don't speak about OTP if the user hasn't enrolled one
On the "password" view in your settings page, the OTP Token fields says: "If
you have enrolled one", should we just show this field if the user has
enrolled one? I figure we have this information since they are displayed in
the previous tab (OTP).

ryanlerch:
previously, the form a logged in user used to change their
password presented fields and descriptions for OTP stuff
when a user hadnt enrolled an OTP token.

This hides those descriptions and fields if a user has not
enabled a OTP token.

Note that the descriptions are still shown on the expired password
reset page as this is not a logged in page, and we cannot know if the
user has an OTP setup there (nor would we want to expose that fact to
the world if not logged in)

Resolves: #206

### DIFF
--- a/noggin/templates/user-settings-password.html
+++ b/noggin/templates/user-settings-password.html
@@ -6,10 +6,12 @@
   <div class="card-body">
     <h5 id="pageheading">{{ _("Change Password") }}</h5>
     <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
-    <div class="form-group">{{ macros.with_errors(password_reset_form.current_password, tabindex="2")}}</div>
+    <div class="form-group" id="currentpasswordinput">{{ macros.with_errors(password_reset_form.current_password, tabindex="2")}}</div>
     <div class="form-group">{{ macros.with_errors(password_reset_form.password, tabindex="3")}}</div>
     <div class="form-group">{{ macros.with_errors(password_reset_form.password_confirm, tabindex="4")}}</div>
-    <div class="form-group">{{ macros.with_errors(password_reset_form.otp, tabindex="5")}}</div>
+    {% if using_otp %}
+    <div class="form-group" id="otpinput">{{ macros.with_errors(password_reset_form.otp, tabindex="5")}}</div>
+    {% endif %}
   </div>
   <div class="card-footer d-flex justify-content-between">
     <div>{{ macros.non_field_errors(password_reset_form) }}</div>
--- a/noggin/tests/unit/controller/test_password_reset.py
+++ b/noggin/tests/unit/controller/test_password_reset.py
@@ -77,6 +77,38 @@ def test_password_changes_user(client, logged_in_dummy_user):
 
 
 @pytest.mark.vcr()
+def test_password_form_with_otp(client, logged_in_dummy_user, dummy_user_with_otp):
+    """Verify that the password change form shows OTP form elements
+       when a user has OTP enabled"""
+    result = client.get("/user/dummy/settings/password")
+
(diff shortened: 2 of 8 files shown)


noggin/templates/user-settings-password.html
```diff
@@ -6,10 +6,13 @@
   <div class="card-body">
     <h5 id="pageheading">{{ _("Change Password") }}</h5>
     <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
-    <div class="form-group">{{ macros.with_errors(password_reset_form.current_password, tabindex="2")}}</div>
+    {# if the user has not enabled OTP, using_otp will be False. If False, we dont show the description in the field that talks about OTP #}
+    <div class="form-group">{{ macros.with_errors(password_reset_form.current_password, tabindex="2", show_description=using_otp)}}</div>
```
abompard: Instead of adding `show_description` and setting it here, you could also change the field description directly in the controller, as in [the doc's example for choice fields](https://wtforms.readthedocs.io/en/stable/fields.html#wtforms.fields.SelectField). It would look something like this in the controller:
```python
if using_otp:
    form.current_password.description = ""
```

abompard (commented):
Just one comment about an alternative way to doing it that seems simpler to me (but maybe not to you). I'll let you decide what to do.

Also, it does not entirely solve #206 because I think the original issue was also pointing out that the password confirmation field for the OTP adding form could be smarter about whether or not there's an OTP already. We could make a pass on the code and check all uses of password fields, see if they are in authenticated pages, and do the same.

ryanlerch:
> Just one comment about an alternative way to doing it that seems simpler to me (but maybe not to you). I'll let you decide what to do.
> 
> Also, it does not entirely solve #206 because I think the original issue was also pointing out that the password confirmation field for the OTP adding form could be smarter about whether or not there's an OTP already. We could make a pass on the code and check all uses of password fields, see if they are in authenticated pages, and do the same.

Yeah, the only other place was the OTP one, and i had already done that one when converting it over to the modal last week:

https://github.com/fedora-infra/noggin/pull/213/commits/eb100b13c3640b62f7ae79855e52d3efe653d6df#diff-244b8263ad1f24aad964c3342e94b84aR49-R53

ryanlerch:
Okies, updated to not do the wierd hide descriiption thing, also becasue we moved that logic into the code, had to write a tests for it, so thats in there now too.

noggin/tests/unit/controller/test_password_reset.py
```diff
@@ -81,6 +81,41 @@ def test_password_changes_user(
...
+
+    page = BeautifulSoup(result.data, "html.parser")
+
+    currentpasswordinput = page.select("#currentpasswordinput .form-text")
+    assert len(currentpasswordinput) == 1
+    assert "Just the password, don't add the OTP token if you have one" in currentpasswordinput[
+        0
+    ].get_text(
+        strip=True
+    )
```
abompard: Well that's a bit verbose (because of black's reformatting), I would have written something like:
```python
currentpasswordinput = page.select_one("#currentpasswordinput .form-text")
assert currentpasswordinput is not None
expected = "Just the password, don't add the OTP token if you have one"
assert expected in currentpasswordinput.get_text(strip=True)
```
And black should leave that alone, I think (not tested).

abompard (changes requested):
Very minor nitpick, sorry... ;-)

abompard (approved):

PR #242, opened 2020-04-22
Add undo buttons to the flash messages for group edits

### LINKED ISSUE #211: Confirmation while adding new users or removing users from a group
While adding new users to a group or removing existing users from a group, it would be nice to give a dialog/popup confirming the change.

ryanlerch:
This adds undo buttons to the flash messages for the actions:

* removing a user from a group
* adding a user to a group

Resolves: #211


abompard (approved):
Works fine, thanks!

PR #244, opened 2020-04-22
Add support for multiple IRC nicks in noggin

### LINKED ISSUE #187: Allow for multiple irc nicks
Some people have multiple irc nicks. Might be nice to allow a list there. Possibly : seperated?

ryanlerch:
fedora-infra/freeipa-fas#82 added support for more than one
IRC Nick to be defined for a user. This PR adds support
for that feature here in Noggin

Resolves: #187


noggin/form/edit_user.py
```diff
@@ -43,7 +45,7 @@ class UserSettingsProfileForm(FlaskForm):
         ],
     )
 
-    ircnick = StringField(_('IRC Nickname'), validators=[Optional()])
+    ircnick = CSVListField(_('IRC Nickname'), validators=[Optional()])
```
StephenCoady:
```suggestion
    ircnick = CSVListField(_('IRC Nicknames'), validators=[Optional()])
```

StephenCoady (changes requested):
One _tiny_ comment!

noggin/controller/user.py
```diff
@@ -76,6 +76,7 @@ def user_settings_profile(ipa, username):
     form = UserSettingsProfileForm(obj=user)
 
     if form.validate_on_submit():
+        app.logger.error(f'******************{form.ircnick.data}*********************')
```
abompard: Looks like a forgotten debug line

noggin/form/base.py
```diff
@@ -65,3 +66,19 @@ class SubmitButtonField(SubmitField):
...
 def strip(value):
     return value.strip() if value else value
+
+
+class CSVListField(Field):
+    widget = TextInput()
+
+    def _value(self):
+        if self.data:
+            return u','.join(self.data)
```
abompard: The `u` prefix is not necessary in python 3

noggin/form/base.py
```diff
@@ -65,3 +66,19 @@ class SubmitButtonField(SubmitField):
...
+        else:
+            return u''
```
abompard: Same here

abompard (changes requested):

noggin/form/base.py
```diff
@@ -65,3 +66,19 @@ class SubmitButtonField(SubmitField):
...
+    widget = TextInput()
+
+    def _value(self):
+        if self.data:
+            return ','.join(self.data)
+        else:
+            return ''
+
+    def process_formdata(self, values):
+        if values:
```
abompard, line 81: I think you should test on `value.strip()`, otherwise if you get a string with only spaces you'll end up with `[""]` instead of an empty list. (feel free to add a unit test for that case ;-) )

abompard (changes requested):
Just one small thing!

noggin/form/edit_user.py
```diff
@@ -43,7 +45,7 @@ class UserSettingsProfileForm(FlaskForm):
         ],
     )
 
-    ircnick = StringField(_('IRC Nickname'), validators=[Optional()])
+    ircnick = CSVListField(_('IRC Nicknames'), validators=[Optional()])
```
abompard, line 48: Maybe add a help text here explaining that this should be comma-separated, otherwise it's not very obvious I think.

ryanlerch: in the UI here we are using selectize, so the user shouldnt ever have to comma separate the values

noggin/form/base.py
```diff
@@ -65,3 +66,19 @@ class SubmitButtonField(SubmitField):
...
+    widget = TextInput()
+
+    def _value(self):
+        if self.data:
+            return ','.join(self.data)
+        else:
+            return ''
+
+    def process_formdata(self, values):
+        if values.strip():
```
abompard: Actually I was wrong, `values` is already a list so it's `values[0]` that should be stripped :-)
And it would be best to do it in the list comprehension below because that list can be empty.

abompard (changes requested):

noggin/form/base.py
ryanlerch: and that is why i shouldnt try to code during meetings, lol.

I think i have it sorted now and should cover all the cases, like `[]` and `[" "]` and `[" , , "]`

also added a new test to check such cases

noggin/tests/unit/controller/test_user.py
```diff
@@ -81,6 +89,20 @@ def test_user_edit_post(client, logged_in_dummy_user):
     )
 
 
+@pytest.mark.vcr()
+def test_user_edit_post_minimal_values(client, logged_in_dummy_user):
+    """Test posting to the user edit page: /user/<username>/settings/profile/
+        with the bare minimum of values """
+    result = client.post('/user/dummy/settings/profile/', data=POST_CONTENTS_MIN)
+    print(result.data)
```
abompard: Looks like a debug leftover... (I usually run a `git grep print` before pushing a PR to check for those, I always forget some)

noggin/tests/unit/form/test_base.py
```diff
@@ -51,3 +51,30 @@ class TestForm(ModestForm):
...
+
+    form = DummyForm()
+
+    # check normal use
+    form.csvfield.process_formdata(["one,two,three"])
+    assert form.csvfield.data == ["one", "two", "three"]
+
+    # check we strip single spaces string, and return []
+    form.csvfield.process_formdata([" "])
+    assert form.csvfield.data == []
```
abompard: You could also use [parametrize](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) to generate test functions independant from each other.

abompard (changes requested):
Just a couple minor things :-)

abompard (approved):

PR #260, opened 2020-04-30
Update user profile

### LINKED ISSUE #212: Not showing all groups that I am part of under "Groups" tab in profile page
I am a member of `ipausers` group and a sponsor of `mohans-group` but in the profile page under groups, I am only seeing `ipausers` group but not the `mohans-group` group.

ryanlerch:
Update the formatting of the user profile page a bit. Mainly to give us a little more room so we can fix #212


StephenCoady:
This looks good to me, I've tested it and its all working OK. Sonarcloud is complaining, but I'm not sure how important the `<i>` stuff is.

abompard:
The `<i>` stuff can be ignored.

PR #262, opened 2020-05-06
escape the OTP token description in the URI

### LINKED ISSUE #241: OTP token description with spaces breaks QR code scanning
If a user creates an OTP token with a description that has spaces in it, the resulting URI (which is encoded into the QR code for scanning into authenticator apps like freeotp and google authenticator) is invalid.

We need to properly encode this value for a URI before returning it.

ryanlerch:
Previously, code was added to noggin that edited the URI we got back
from freeipa, adding the description of the token to the OTP URI. this
was done, so the token description is shown in the authenticator
application that the user uses. However, if a user added a token
description with a space in it,  the URI would be constructed to be
invalid, and not be able to be scanned via the QR code.

This PR escapes the description before adding it to the otp URI, which
resolved this issue.

Resolves: #241

### DIFF
--- a/noggin/controller/user.py
+++ b/noggin/controller/user.py
@@ -1,4 +1,4 @@
-from urllib.parse import urlparse
+from urllib.parse import urlparse, quote
 
 from flask import flash, redirect, render_template, session, url_for, Markup
 from flask_babel import _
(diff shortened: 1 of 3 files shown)


StephenCoady (approved):
looks good to me, tested and working.

PR #263, opened 2020-05-07
Add email blocklist validator

### LINKED ISSUE #234: blacklisting some domains when registering account (to avoid loop with aliases)
In the actual deployed instance, noggin let you register with any email address (there is another ticket open for validation)
Problem is that it doesn't refuse some domains, including @fedoraproject.org .
But because email aliases are generated from FAS/IPA, users shouldn't be able to register with such alias (generated *from* IPA), as that would cause a loop (even for password reset instructions, etc)

ryanlerch:
Add extra validation on email fields to read a new validator blacklist
from the noggin configuration, and not let users change their email
addresses to any email domain in the blacklist.

Resolves: #234

### DIFF
--- a/noggin/form/validators.py
+++ b/noggin/form/validators.py
@@ -1,0 +1,19 @@
+from flask_babel import lazy_gettext as _
+from wtforms.validators import ValidationError
+from wtforms.validators import Email as WTFormsEmailValidator
+
+
+class Email(WTFormsEmailValidator):
+    """ Extend the WTForms email validator, adding the ability to blocklist
+        email addresses
+    """
+
+    def __init__(self, blocklist=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.blocklist = blocklist
+
+    def __call__(self, form, field):
+        super().__call__(form, field)
+        domain = field.data.split("@")[1]
+        if domain in self.blocklist:
+            raise ValidationError(_('Email addresses from that domain are not allowed'))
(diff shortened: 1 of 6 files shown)


abompard (changes requested):
If you don't mind, I'd prefer if you changed all mentions of "blacklist" to "blocklist". I've heard that the blacklist / whitelist expressions could be interpreted as racially connoted (white = good, black = bad). We don't think about it because it's so common in the software industry (and also maybe because this team is only composed only of white people at the moment...) but I think it's still a valid concern. And it's really trivial to replace with block/allow :-)

Conan-Kudo (approved):

noggin/form/validators.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def __init__(self, blocklist=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.blocklist = blocklist
+
+    def __call__(self, form, field):
+        try:
+            super().__call__(form, field)
+        except ValidationError as e:
+            raise StopValidation(e)
```
abompard: I don't think it's technically necessary to do this, unless you want validation to stop for validators that would be added *after* the Email validation class in the field's `validators` argument. If the only thing you want is to avoid the `IndexError` caused by the absence of a `@` in the field value, then you can let the `ValidationError` bubble up from the `super()` call.
Both behaviors are fine for me.

abompard (commented):
Just a small comment to make sure what you wanted to accomplish. Looks good in general, thanks!

abompard:
There's still a minor linting issue but apart from that it looks good, thanks!

ryanlerch:
fixed the linting issue! should be right to go when the tests pass

abompard (approved):

PR #266, opened 2020-05-11
Fix lowercase bug with username

### LINKED ISSUE #255: can log in with username in different cases
currently, it is possible to log in to noggin with my username in a different case.

For example, i can log in with the username, 'megan0' and the welcome flash message states:

```
Welcome, megan0!
```
and i can access the user settings page.

However, if i log into the same user with the username as a different case, for example: "meGan0", the welcome flash message states:

```
Welcome, meGan0!
```
and if i try to access my user settings, it says i am not permitted to.

james02135:
fixes #255

### DIFF
--- a/noggin/tests/unit/controller/test_authentication.py
+++ b/noggin/tests/unit/controller/test_authentication.py
@@ -109,6 +109,44 @@ def test_login_no_username(client):
     assert "noggin_username" not in session
 
 
+def test_login_username_case(client, dummy_user):
+    """Test giving username random uppercase letter"""
+    result = client.post(
+        '/',
+        data={
+            "login-username": "duMmy",
+            "login-password": "dummy_password",
+            "login-submit": "1",
+        },
+        follow_redirects=True,
+    )
+    page = BeautifulSoup(result.data, 'html.parser')
+    messages = page.select(".flash-messages .alert-success")
+    assert len(messages) == 1
+    assert messages[0].get_text(strip=True) == 'Welcome, dummy!×'
+    assert session.get("noggin_username") == "dummy"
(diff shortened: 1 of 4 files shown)


noggin/tests/unit/controller/test_authentication.py
```diff
@@ -109,6 +109,44 @@ def test_login_no_username(client):
...
+    page = BeautifulSoup(result.data, 'html.parser')
+    messages = page.select(".flash-messages .alert-success")
+    assert len(messages) == 1
+    assert messages[0].get_text(strip=True) == 'Welcome, dummy!×'
+    assert session.get("noggin_username") == "dummy"
+    assert session.get("noggin_session") is not None
+    
+
+def test_login_username_created_with_case(client, dummy_user_with_case):
+    """Test giving username random uppercase letter"""
```
StephenCoady: This should be changed to reflect the test, looks like it was copied from the other test.

StephenCoady (changes requested):
One small comment, also you need to sign your commits for the dco.

StephenCoady:
also, if you add the text

```
fixes #255
```
to your description it will link the issue automatically. me doing it here may do that too, but i'm not sure. lets see.

EDIT: nope. add it to the description.

StephenCoady:
So when you add new commits, the original commits are left untouched. Lets squash these 4 commits into 1 and then ammend that one commit so that it has the correct signoff.

* git rebase -i <commit hash before your first commit>
* fix the 3 follow up commits you made so that they are squashed
* git commit --amend --signoff
* git push

StephenCoady (approved):
good work!

PR #272, opened 2020-05-12
Client Migration to python-freeipa 1.0

### LINKED ISSUE #94: As a developer, I want to migrate to freeipa client library 1.0.0
We are currently using the [python-freeipa](https://python-freeipa.readthedocs.io) library ([source code](https://github.com/opennode/python-freeipa)), which is not the official client library and has recently been updating to version 1.0.0, breaking compatibility with what we were using. They do provide a legacy module with a compatible class, so we can keep using it for now, but before we adapt to the non-legacy code we should decide whether we want to keep using that library or move to the official [freeipa client library](https://github.com/freeipa/freeipa/tree/master/ipaclient), which is also used to power the `ipa` command line, and as a result can do all operations that the command line provides. The current client is lacking some operations and we've had to extend it.
However, @relrod said that the official library is harder to use, and we may need help from the FreeIPA team if we decide to use it. I think we should investigate those options.

james02135:
Signed-off-by: james02135 <james02135@hotmail.com>


james02135:
first batch of major changes, still a WIP but would appreciate a review

noggin/controller/registration.py
```diff
@@ -65,33 +65,33 @@ def _handle_registration_validation_error(username, e):
 
 
 def handle_register_form(form):
-    username = form.username.data
+    a_uid = form.username.data
```
StephenCoady: I think it would be better to leave this as username, its more readable in the rest of the method.

noggin/security/ipa_admin.py
```diff
@@ -6,30 +6,6 @@
 
 class IPAAdmin(object):
 
-    __WRAPPED_METHODS = (
```
StephenCoady: these methods will still need to be wrapped, because they are performed by the admin account.

noggin/tests/unit/conftest.py
```diff
@@ -102,27 +100,27 @@ def ipa_testing_config(vcr_session):
 def make_user(ipa_testing_config):
     created_users = []
 
-    def _make_user(name):
+    def _make_user(a_uid):
```
StephenCoady: same comment here as before. you can leave the function the same but just assign `a_uid` if you like.

noggin/tests/unit/conftest.py
```diff
@@ -171,31 +169,31 @@ def logged_in_dummy_user(client, dummy_user):
...
+        o_ipatokenowner="dummy",
+        o_ipatokenotpalgorithm='sha512',
+        o_description="dummy's token",
     )
     token = OTPToken(result)
     yield token
     # Deletion needs to be done as admin to remove the last token
     try:
-        ipa_admin.otptoken_del(token.uniqueid)
+        ipa.otptoken_del(a_ipatokenuniqueid=token)
```
StephenCoady: this will still need to be done by the admin

noggin/tests/unit/conftest.py
```diff
@@ -171,31 +169,31 @@ def logged_in_dummy_user(client, dummy_user):
...
     except python_freeipa.exceptions.NotFound:
         pass  # Already deleted
 
 
 @pytest.fixture
-def cleanup_dummy_tokens():
+def cleanup_dummy_tokens(client):
```
StephenCoady: do you need this?

noggin/utility/__init__.py
```diff
@@ -73,7 +73,7 @@ def group_or_404(ipa, groupname):
 
 def user_or_404(ipa, username):
     try:
-        return ipa.user_show(username)
+        return ipa.user_find(o_uid=username)
```
StephenCoady: any reason in particular to move from `user_show`?

StephenCoady: any reason in particular to move from `user_show`?

StephenCoady (changes requested):
Few comments. The main one being we will still need to wrap certain methods in the `ipa_admin.py` file. we don't need to explicitly have them in the `ipa.py` file anymore, but they will need to be wrapped.

StephenCoady:
@james02135 this will need to be rebased as you have conflicts. also, most of the test suite is failing. if you run `tox` inside your vm it will execute all the runners for you.

nphilipp:
@james02135, added the WIP label because this needs some work: rebasing on top of `dev` (see below), a better commit log message and test fixes.

As I can't push to your fork, I've rebased on top of current `dev` and resolved conflicts in my fork here: https://github.com/nphilipp/noggin/tree/dev--pr272-ClientMeta

This branch also has a commit on top which reverts changes to `poetry.lock` which introduce unneeded (and in part unfulfillable) new dependencies, I assume this slipped in inadvertently.

nphilipp:
The latest push restores coverage:

* It removes the `parse_group_management_error()` function from `noggin.security.ipa` which was only used in methods which were migrated upstream into `python-freeipa`.
* It marks one code path as `pragma: no cover` because it would be covered by this test which we expect to fail.

noggin/tests/unit/conftest.py
```diff
@@ -87,7 +87,6 @@ def ipa_testing_config(vcr_session):
         except python_freeipa.exceptions.BadRequest as e:
             if not e.message == "no modifications to be performed":
                 raise
-            raise
```
abompard, line 90: Oh yeah I think that's my fault, I've put this in as a debug line and forgot to take it off

noggin/controller/password.py
```diff
@@ -91,7 +91,14 @@ def user_settings_password(ipa, username):
     form = PasswordResetForm()
 
     # check if an OTP token exists. If so, the user is using OTP.
-    using_otp = bool(ipa.otptoken_find(ipatokenowner=username))
+    using_otp = bool(
+        ipa.otptoken_find(
+            o_ipatokenowner=username,
+            o_ipatokenotpalgorithm='sha512',
```
abompard: This line and the next 2 should not be necessary with python_freeipa >= 1.0.6 (see [this PR](https://github.com/opennode/python-freeipa/pull/44)), we can leave it here but could you please add a comment in the code to explain? Thanks.

noggin/controller/user.py
```diff
@@ -188,7 +193,15 @@ def user_settings_otp(ipa, username):
     otp_uri = session.get('otp_uri')
     session['otp_uri'] = None
 
-    tokens = [OTPToken(t) for t in ipa.otptoken_find(ipatokenowner=username)]
+    tokens = [
+        OTPToken(t)
+        for t in ipa.otptoken_find(
+            o_ipatokenowner=username,
+            o_ipatokenotpalgorithm='sha512',
```
abompard: Same remark as above

noggin/controller/user.py
```diff
@@ -271,7 +284,7 @@ def user_settings_otp_delete(ipa, username):
         username = session.get('noggin_username')
         token = form.token.data
         try:
-            ipa.otptoken_del(ipatokenuniqueid=token)
+            ipa.otptoken_del(a_ipatokenuniqueid=token)['result']
```
abompard: Seems weird to access the `result` key when we're not storing it in a variable.

noggin/security/ipa_admin.py
```diff
@@ -10,6 +10,7 @@ class IPAAdmin(object):
         "user_add",
         "user_show",
         "user_mod",
+        "user_del",
```
abompard: Where is this used? I don't think it should be necessary, noggin does not (and should not be allowed to) delete users. If it's only in the tests it should be in the `__WRAPPED_METHODS_TESTING` tuple below (it's already there).

noggin/tests/unit/conftest.py
```diff
@@ -178,34 +180,33 @@ def logged_in_dummy_user(client, dummy_user):
...
+        o_ipatokenotpalgorithm='sha512',
+        o_description="dummy's token",
     )
-    token = OTPToken(result)
+    token = OTPToken(result['result'])
     yield token
     # Deletion needs to be done as admin to remove the last token
     try:
-        ipa_admin.otptoken_del(token.uniqueid)
+        ipa_admin.otptoken_del(result['result']['dn'][0])
```
abompard: Are you sure? In `ClientMeta` it looks like the argument is the `uniqueid`, not the `dn`.

noggin/tests/unit/controller/test_group.py
```diff
@@ -241,13 +241,6 @@ def test_group_remove_member_hidden_group(
 @pytest.mark.vcr()
 def test_group_remove_self(client, logged_in_dummy_user, dummy_group):
     """Test a non-sponsor user removing themselves from a group"""
-    ipa_admin.group_add_member("dummy-group", users="dummy")
-    result = client.get('/group/dummy-group/')
-    assert result.status_code == 200
-    page = BeautifulSoup(result.data, 'html.parser')
-    leave_btn = page.select_one("#leave-group-btn")
-    assert leave_btn.get_text(strip=True) == "Leave group"
```
abompard: Why remove all that?

noggin/tests/unit/controller/test_user_otp.py
```diff
@@ -20,14 +20,14 @@
...
-    token = OTPToken(result)
+        a_ipatokenuniqueid=None,
+        o_ipatokenowner="dummy",
+        o_ipatokenotpalgorithm="sha512",
+        o_description="dummy's other token",
+    )['result']
     yield (dummy_user_with_otp, OTPToken(result))
     try:
-        ipa_admin.otptoken_del(token.uniqueid)
+        ipa_admin.otptoken_del(result['dn'][0])
```
abompard: Same here, it looks like `otptoken_del` takes the uniqueid as argument, not the dn.

noggin/tests/unit/controller/test_user_otp.py
```diff
@@ -525,23 +525,23 @@ def test_user_settings_otp_enable(client, logged_in_dummy_user, dummy_user_with_
     # grab the id of the first token
     tokenid = tokenlist[0].select(".text-monospace")[0].get_text(strip=True)
 
-    # disable that token
+    # delete that token
     result = client.post(
-        "/user/dummy/settings/otp/disable/",
+        "/user/dummy/settings/otp/delete/",
```
abompard: I don't understand this change and the rest in the test, it looks like you're testing something else.

abompard (changes requested):

noggin/controller/password.py
nphilipp: Removing these lines with `python-freeipa` 1.0.6 breaks tests. This is with a reset & repopulated FreeIPA and removed VCR cassettes (which works without the removed lines).

noggin/controller/user.py
nphilipp: Yeah, that can go away.

noggin/security/ipa_admin.py
nphilipp: Probably just a case of cargo-culting. :wink:

noggin/controller/password.py
nphilipp: I destroyed and re-provisioned my Vagrant box, and now it works. No idea why, though.

noggin/tests/unit/controller/test_group.py
nphilipp: That was a mistake and we added it back after reviewing it on Thursday.

noggin/security/ipa_admin.py
nphilipp: Removing this, and the added `pwpolicy_find` because it's not even used in testing.

noggin/tests/unit/conftest.py
nphilipp: Agreed.

noggin/tests/unit/controller/test_user_otp.py
nphilipp: The new `ClientMeta` seems to have changed the semantics of `otptoken_mod()` slightly, i.e. setting `o_ipatokendisabled=None` won't reset it to the default (`False`) anymore, which broke enabling disabled OTP tokens, thus the modifications to the test (deleting a token worked :wink:). I've fixed how the controller method invokes `otptoken_mod()` and reverted the changes to the test in the latest push.

noggin/tests/unit/controller/test_user_otp.py
```diff
@@ -256,7 +257,7 @@ def test_user_settings_otp_disable_ipaerror(
         )
         result = client.post(
             "/user/dummy/settings/otp/disable/",
-            data={"token": dummy_user_with_2_otp[1].uniqueid},
+            data={"token": dummy_user_with_2_otp[1]},
```
abompard: Uh, why this change?

noggin/controller/user.py
```diff
@@ -31,13 +31,15 @@ def user(ipa, username):
...
+            a_criteria=None, o_user=username, o_all=False, fasgroup=True
+        )['result']
     ]
     managed_groups = [
         Group(g)
-        for g in ipa.group_find(membermanager_user=username, all=False, fasgroup=True)[
-            'result'
-        ]
+        for g in ipa.group_find(
+            a_criteria=None, o_membermanager_user=username, o_all=False, fasgroup=True
```
abompard: It's not necessary to specify `a_criteria=None` as it's the default.

noggin/tests/unit/controller/test_user_otp.py
nphilipp: I'm not sure about this one myself. I'll give it a go with that change backed out.

noggin/controller/user.py
nphilipp: That makes sense.

nphilipp:
This latest push contains the discussed changes and doesn't hardcode `a_ipatokenuniqueid="dummy"` anymore in `noggin/tests/unit/conftest.py`.

abompard (approved):
:fireworks:

PR #275, opened 2020-05-14
Re-word the email templates for forgot password

### LINKED ISSUE #178: The email sent to recover a user's password could be improved
Currently the password sent with the link to recover a user's password is pretty poorly formatted, we could make it look way better.

One important thing to keep in mind is to do our best to prevent these emails from being identified as spam by mail servers/filters.

ryanlerch:
Resolves: #178


abompard (approved):
Thanks

Issue #209, opened 2020-04-07
As an application owner, I want a fedora-messaging message to be emitted when some user data are changed


mscherer:
We need the noggin to emit a message when a user change their ssh key, timezone, gpg key or password to emit a message, because the Badges application uses it. FAS used to emit a message on the fas.user.update topic, the contents could be modelled after it (or adapted).

**Acceptance Criteria**:
* a Fedora Messaging message is emitted when a user is updated
* to ease transition and if it's possible, the message should be emitted on the same topic with the same content.
* a crash in the message sending code should cause a message to be logged with full traceback, but the page should not fail rendering.

**Definition of Done**:
* all the above is done
* unit tests for 100% coverage (unless explicitely ignored with an explanation)
* CI passes
* the path to the Fedora Messaging configuration can be set by a configuration variable.
* the message should use schemas
* there is documentation on the message schema and its contents

abompard:
There is controversy around publishing a public message when a users changes their password (see https://github.com/fedora-infra/noggin/issues/25#issuecomment-582961135). Can the applications do without that notification?

mscherer:
No, we give badge based on that precise event:
https://badges.fedoraproject.org/badge/paranoid-panda

I would even say that we could use badge as a incentive to change password on a regular basis, which could be a good idea too (even if mandatory rotation is not great)

I would also add that seeing that a password was changed could also tell something fishy did happen. Like if I know that a admin X is in vacation, but there is a password change, maybe there is something fishy.

And as indicated in the other ticket, I think a clearer rational than "I hate this" could be provided, as there isn't much discussion to have on that fact.

abompard:
@relrod , what do you think of this?

relrod:
Password changes should pretty clearly not be a public event. Consider a scenario where someone with a high level of access is being targeted. Broadcasting this information tells the attacker "restart your search" which is information we need to not leak.

There's a reason this information is kept in shadow and not passwd on Linux systems and why reading the journal (where password changes are logged) requires elevated privileges.

```
[codeblock@batcave01 ~][PROD]$ chage -l root
chage: Permission denied.
```

mscherer:
I disagree on the reasoning.

[...]

PR #284, opened 2020-05-28
add messaging for password changes

ryanlerch:
This adds support for fedora messages when the user changes their password.

Note that fedora-infra/noggin-messages#15 will need to be merged before the CI will pass

Related: #209 


abompard (approved):

relrod:
I still think this is a really bad idea.

nbebout:
I'm not sure that we need to publish a message when a password changes.  Could this possibly be misused? I'm not sure.

PR #285, opened 2020-05-29
Send a message when the user updates their profile

### LINKED ISSUE #209: As an application owner, I want a fedora-messaging message to be emitted when some user data are changed

ryanlerch:
With this change, a message is now sent via Fedora Messaging that
contains the names of the feilds that a user has changed when they
update their profile

Resolves: #209


noggin/representation/user.py
```diff
@@ -67,3 +67,24 @@ def rhbz_mail(self):
...
+        }
+
+    def diff_against(self, other):
+        """
+        Compares two instances of the User class, and returns the properties
+        with values that are different
+        """
+        return [
+            key
+            for key, value in self.get_properties().items()
```
abompard: Why not use the `.raw` attribute? And do something like `for key, value in self.raw.items()`

noggin/representation/user.py
```diff
@@ -67,3 +67,24 @@ def rhbz_mail(self):
...
+            if value != other.get_properties()[key]
```
abompard: same here, you can compare `value` to `other.raw[key]`. I don't think you need the `get_properties()` method.

noggin/representation/user.py
```diff
@@ -67,3 +67,24 @@ def rhbz_mail(self):
...
+            key: getattr(self, key)
+            for key, value in self.__class__.__dict__.items()
+            if isinstance(value, property)
+        }
+
+    def diff_against(self, other):
+        """
+        Compares two instances of the User class, and returns the properties
+        with values that are different
+        """
```
abompard: You could move that method to the parent `Representation` class, so it's available to all our entity representations.
Also, please add a check on the nature of the object that we're diffing with, for example:
```python
if not isinstance(other, self.__class__):
    raise ValueError(f"Can't diff a {self.__class__.__name__} instance against a {other.__class__.__name__} instance")
```

abompard (changes requested):

noggin/controller/user.py
```diff
@@ -52,26 +54,40 @@ def user(ipa, username):
...
         flash(
             Markup(
-                f'Profile Updated: <a href=\"{url_for("user", username=username)}\">'
+                f'Profile Updated: <a href=\"{url_for("user", username=user.username)}\">'
                 'view your profile</a>'
             ),
             'success',
         )
-        return redirect(url_for(redirect_to, username=username))
+        updated_user = User(user_or_404(ipa, user.username))
```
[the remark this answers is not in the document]
ryanlerch: for some reason, the ipa.user_mod seems to return some different values -- always getting the last_password_change as a changed field here. that is why i resorted to another api call here.

noggin/representation/user.py
ryanlerch: .raw contains the ipa names, i.e. "uid" rather than "username",that is why i added get_properties()

noggin/representation/user.py
ryanlerch: .raw contains the ipa names, i.e. "uid" rather than "username",that is why i added get_properties()

noggin/representation/user.py
ryanlerch: Done, and re-formatted the tests a bit so i could test this.

noggin/controller/user.py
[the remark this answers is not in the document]
abompard: Huh, that's wierd, I'll investigate this.

noggin/representation/user.py
abompard: Ah, I see. It still feels a bit "unnatural" to do it this way, I'll try another way, maybe using `__iter__`.

noggin/controller/user.py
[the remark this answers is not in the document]
abompard: OK I found why. The `user_mod` call  was not returning all fields, so the `last_password`change` was showing up as changed because it was empty in the new object.
I've added the `all=True` parameter to the call so we get all fields, and it worked.

ryanlerch:
@abompard i am happy with your changes, are we ok to merge?

abompard (approved):

PR #289, opened 2020-06-04
Update documentation

### LINKED ISSUE #246: noggin vs fas2 quick reference
It would be nice to have a small user oriented document about the differences between fas2 and noggin. for example, in fas2 groups have admins that can make users into sponsors and when using noggin such a change needs a ipa admin to make the change for you. 

This would be something we could point users that are used to fas2 to more easily move to noggin.

ryanlerch:
Resolves: #246 

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -103,6 +103,7 @@ extlinks = {
 
 # -- Misc -----
 
+
 def run_apidoc(_):
     from sphinx.ext import apidoc
 
@@ -112,6 +113,7 @@ def run_apidoc(_):
             "-o",
             os.path.join(topdir, "docs", "_source"),
             os.path.join(topdir, "noggin"),
+            os.path.join(topdir, "noggin", "tests"),
         ]
     )
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -100,3 +100,21 @@ extlinks = {
     "issue": ("https://github.com/fedora-infra/noggin/issues/%s", "#"),
     "pr": ("https://github.com/fedora-infra/noggin/pull/%s", "PR#"),
 }
+
+# -- Misc -----
+
+def run_apidoc(_):
+    from sphinx.ext import apidoc
+
+    apidoc.main(
+        [
+            "-f",
+            "-o",
+            os.path.join(topdir, "docs", "_source"),
+            os.path.join(topdir, "noggin"),
+        ]
+    )
+
+
+def setup(app):
+    app.connect("builder-inited", run_apidoc)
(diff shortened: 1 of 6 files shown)


docs/conf.py
```diff
@@ -100,3 +100,22 @@
...
+
+def run_apidoc(_):
+    from sphinx.ext import apidoc
+
+    apidoc.main(
+        [
+            "-f",
+            "-o",
+            os.path.join(topdir, "docs", "_source"),
+            os.path.join(topdir, "noggin"),
```
abompard, line 115: We can safely exclude the tests, just add `noggin/tests` to the arguments list.

abompard (changes requested):
While investigating why it was not producing documentation for the controllers, I realized that we indeed do *not* have a `noggin/controller/__init__.py` file. Incredible that it works at all. Please add it (empty is fine).

docs/conf.py
```diff
@@ -100,3 +100,21 @@
     "issue": ("https://github.com/fedora-infra/noggin/issues/%s", "#"),
     "pr": ("https://github.com/fedora-infra/noggin/pull/%s", "PR#"),
 }
+
```
abompard, line 103: CI is unhappy because you need one more blank line here I think. Just run black on the file and you'll be all set

abompard (changes requested):

docs/conf.py
ryanlerch: not sure what you mean by this comment, can you clarify for me?

abompard: Sure. As it is now, you are generating the apidocs for the unit tests modules, and I don't think it's useful. If you add `noggin/tests` to the command line, it'll exclude them from apidoc generation.

ryanlerch: thanks for that!

fixed and updated!

docs/index.rst
```diff
@@ -28,8 +24,22 @@ Here is what works so far:
 
 Migrating your application:
 
-If your application currently uses FAS, and you wish to migrate it to use Noggin, please take a
-look at FASJSON's documentation which can be found `here <https://fasjson.readthedocs.io/>`_.
+If your application currently uses communicates with the Fedora Account System (FAS2) 
+using it's readonly API, this functionality will be provided by `fasjson <https://github.com/fedora-infra/fasjson/>`_
```
abompard: Missing full-stop! Sorry! :-)

abompard (changes requested):

abompard (approved):

PR #291, opened 2020-06-04
Pagination for the groups page

### LINKED ISSUE #279: Paginated queries from noggin to FreeIPA
Right now the list of groups is queried entirely from FreeIPA, I would be very useful to be able to do paged queries so we can display groups in pages.

abompard:
This PR adds a pagination system and makes the groups page paginated.

Since we can't set limits and offsets in the IPA queries, this uses the same system that we use with LDAP in FASJSON: 

1. request the entire list of primary keys (in the case of groups, their names),
2. build the list of groups for the current page
3. query details for those groups in batch (one API call)

This PR also contains a couple or minor fixes in separate commits.


ryanlerch (approved):

abompard:
CI is broken at the moment, merging manually.

PR #299, opened 2020-06-11
Add the new group attributes

### LINKED ISSUE #294: Groups have attributes now, noggin should display them.
Since fedora-infra/freeipa-fas#104 groups have a few attributes. They should be displayed in Noggin.

ryanlerch:
3 new group attributes were added to freeipa-fas. this adds
these attributes (mailing list, url, and ircchannel) to noggin

Resolves: #294

### DIFF
--- a/noggin/representation/group.py
+++ b/noggin/representation/group.py
@@ -8,8 +8,11 @@ class Group(Representation):
         "description": "description",
         "members": "member_user",
         "sponsors": "membermanager_user",
+        "url": "fasurl",
+        "irc_channel": "fasircchannel",
+        "mailing_list": "fasmailinglist",
     }
-    ATTR_LISTS = ["members", "sponsors"]
+    ATTR_LISTS = ["members", "sponsors", "url"]
 
     pkey = "name"
     ipa_object = "group"
--- a/noggin/representation/group.py
+++ b/noggin/representation/group.py
@@ -8,11 +8,11 @@ class Group(Representation):
         "description": "description",
         "members": "member_user",
         "sponsors": "membermanager_user",
-        "url": "fasurl",
+        "urls": "fasurl",
         "irc_channel": "fasircchannel",
         "mailing_list": "fasmailinglist",
     }
-    ATTR_LISTS = ["members", "sponsors", "url"]
+    ATTR_LISTS = ["members", "sponsors", "urls"]
 
     pkey = "name"
     ipa_object = "group"
--- a/noggin/templates/group.html
+++ b/noggin/templates/group.html
@@ -10,25 +10,38 @@
   <div class="container section">
     <div class="row py-4" data-section="identity">
       <div class="col">
-          <h3>{{ group.name }}</h3>
+          <h3 class="mb-0">
+            {{ group.name }}
+            {% if edit_group_details is defined %}
+              <span class="pl-1">{{ edit_group_details()}}</span>
+            {% endif %}
+          </h3>
           {% if group.description %}<div>{{ group.description }}</div>{% endif %}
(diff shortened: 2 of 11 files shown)


noggin/templates/group.html
```diff
@@ -10,25 +10,38 @@
...
           {% if group.description %}<div>{{ group.description }}</div>{% endif %}
+          {% if group.mailing_list %}
+            <div><i class="fa fa-fw fa-envelope" aria-hidden="true"></i> <a href="{{group.mailing_list}}">{{group.mailing_list}}</a></div>
+          {% endif %}
+          {% if group.irc_channel %}
+            <div><i class="fa fa-fw fa-comments-o" aria-hidden="true"></i> <a href="{{group.irc_channel}}">{{group.irc_channel}}</a></div>
+          {% endif %}
+      </div>
+      <div class="col-auto align-self-center">
+          {% if group.url %}
```
abompard: You don't actually need the `if` block, if the list it empty the `for` loop will just produce no output. Unless you want to put the `if` block around the `div`.

noggin/representation/group.py
```diff
@@ -8,8 +8,11 @@ class Group(Representation):
         "description": "description",
         "members": "member_user",
         "sponsors": "membermanager_user",
+        "url": "fasurl",
```
abompard: Since it's multi-valued, maybe call it `urls`? We've done that for other values.

noggin/themes/fas/templates/main.html
```diff
@@ -96,13 +96,15 @@ <h2 class="display-4">{{_("Fedora Accounts")}}</h2>
...
 {# typically, it used to link to a ticket tracker for an admin to edit a group #}
 {% macro edit_group_details() %}
-    {% if current_user_is_sponsor and sponsor_form %}
-        <div class="col-auto ml-auto">
-            <a class="btn btn-sm btn-outline-primary" href="https://pagure.io/fedora-infrastructure/new_issue" title="{{_('file a Fedora Infra ticket to change the details or sponsors of this group')}}">
-                {{_("Request Change of Details")}}
-            </a>
-        </div>
-    {% endif %}
+    <a tabindex="0"
```
abompard: To avoid the `href=javascript` stuff, you can use a `button` tag with `btn-link` in the classes.

abompard (changes requested):

noggin/themes/fas/templates/main.html
```diff
@@ -96,13 +96,15 @@ <h2 class="display-4">{{_("Fedora Accounts")}}</h2>
...
-                {{_("Request Change of Details")}}
-            </a>
-        </div>
-    {% endif %}
+    <button tabindex="0"
+        class="btn btn-link"
+        role="button" data-toggle="popover" 
+        data-trigger="focus" 
+        title="Edit Group" 
+        data-content="To change group details or add sponsors, file a ticket with <a href='https://pagure.io/fedora-infrastructure/new_issue' target='_blank'>fedora-infra</a>"
```
abompard: This needs to be flagged for translation.

noggin/themes/fas/templates/main.html
```diff
@@ -73,16 +73,16 @@
     {{ super() }}
 {% endblock %}
 
-{% block userlinks %}
+{% macro userlinks() %}
```
abompard, line 76: It's still a `block` in other themes, please change them too.

abompard (changes requested):
Could you also update the `noggin.tests.unit.controller.test_group:test_group()` test to check for the new attributes? Thanks!

abompard (approved):

PR #312, opened 2020-07-10
Rebuild cass

ryanlerch:
Two commits in this one,

the first is a change to the vagrant environment that takes a snapshot of the empty IPA data, and adds a few extra aliased commands to reset the data, and also repopulate with the test data.

Fixed a bunch of the tests that broke when rebuilding the cassettes, and also rebuilt all the cassettes. Just note here that the cassettes that are checked in were all rebuilt with the empty FreeIPA data, but rebuilding them all with the test data too works.

### DIFF
--- a/noggin/controller/user.py
+++ b/noggin/controller/user.py
@@ -301,7 +301,7 @@ def user_settings_otp_delete(ipa, username):
 def user_settings_agreements(ipa, username):
     user = User(user_or_404(ipa, username))
     agreements = [
-        Agreement(a) for a in ipa.fasagreement_find(all=False) if Agreement(a).enabled
+        Agreement(a) for a in ipa.fasagreement_find(all=False, ipaenabledflag=True)
     ]
     form = UserSettingsAgreementSign()
     if form.validate_on_submit():
--- a/noggin/tests/unit/controller/test_user_otp.py
+++ b/noggin/tests/unit/controller/test_user_otp.py
@@ -1,6 +1,7 @@
 import python_freeipa
 import mock
 import pytest
+
 from bs4 import BeautifulSoup
 
 from noggin import ipa_admin
@@ -157,10 +158,10 @@ def test_user_settings_otp_check_description_escaping(
     page = BeautifulSoup(result.data, "html.parser")
     otp_uri = page.select_one("input#otp-uri")
 
+    # the escaping happens before the secret, so just check that
     assert (
-        otp_uri['value'] == "otpauth://totp/dummy@example.com:pants%20token?issuer="
-        "dummy%40EXAMPLE.COM&secret=L4PD6EXABBJDSCAKS6MZQWT4RSP3PM3QW6H57UHIKFCN7I3"
-        "FGKSHZCCO&digits=6&algorithm=SHA512&period=30"
+        otp_uri['value'][:81]
+        == "otpauth://totp/dummy@example.com:pants%20token?issuer=dummy%40EXAMPLE.COM&secret="
     )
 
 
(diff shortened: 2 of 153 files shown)


noggin/controller/user.py
```diff
@@ -301,7 +301,7 @@ def user_settings_otp_delete(ipa, username):
 def user_settings_agreements(ipa, username):
     user = User(user_or_404(ipa, username))
     agreements = [
-        Agreement(a) for a in ipa.fasagreement_find(all=False) if Agreement(a).enabled
+        Agreement(a) for a in ipa.fasagreement_find(all=True) if Agreement(a).enabled
```
ryanlerch: with all=False here, the enabled boolean wasnt being sent back from the fasagreement_find call, so no agreements were not showing up in the agreements list, and after rebuilding the cassettes, theses tests were failing, but changing to all=true fixed the issue

noggin/tests/unit/controller/test_user_otp.py
```diff
@@ -157,10 +158,10 @@ def test_user_settings_otp_check_description_escaping(
     page = BeautifulSoup(result.data, "html.parser")
     otp_uri = page.select_one("input#otp-uri")
 
+    # the escaping happens before the secret, so just check that
     assert (
-        otp_uri['value'] == "otpauth://totp/dummy@example.com:pants%20token?issuer="
-        "dummy%40EXAMPLE.COM&secret=L4PD6EXABBJDSCAKS6MZQWT4RSP3PM3QW6H57UHIKFCN7I3"
-        "FGKSHZCCO&digits=6&algorithm=SHA512&period=30"
+        otp_uri['value'][:81]
+        == "otpauth://totp/dummy@example.com:pants%20token?issuer=dummy%40EXAMPLE.COM&secret="
```
ryanlerch, line 164: hardcoding the secret here was causing issues. b ut because we are just testing for the escaping of the token name and the issuer, I truncated the string before the secret hash so it should work across cassette rebuilds.

noggin/controller/user.py
abompard: Doesn't it return all the member lists too? Maybe we should change the freeipa-fas code to return the enabled boolean as well.

ryanlerch: what do you think about changing to this?
```
    agreements = [
        Agreement(a) for a in ipa.fasagreement_find(all=False, ipaenabledflag=True)
    ]
```

abompard: Oh of course, that's much better!

ryanlerch: fixed!

abompard (approved):

noggin/tests/unit/controller/test_user_otp.py
nphilipp: As I understand it, we're mainly interested in that the `description` part in the URI is escaped.

Instead of dealing with the vagaries of how the URI is constructed (I could imagine that order of query params isn't terribly stable), how about we use `urllib.parse.urlparse()` and only check the parts we care about? E.g.:

````
from urllib.parse import parse_qs, urlparse
...
    otp_uri = page.select_one("input#otp-uri")
    parsed_otp_uri = urlparse(otp_uri)
    parsed_query = parse_qs(parsed_otp_uri.query)

    # Not sure we need all of these checked
    assert parsed_otp_uri.scheme == "otpauth"
    assert parsed_otp_uri.netloc == "totp"
    assert parsed_otp_uri.path == "/dummy@example.com:pants%20token"

    assert parsed_query["issuer"] == "dummy@EXAMPLE.COM"
````

nphilipp (approved):

PR #314, opened 2020-07-14
Miscellaneous fixes...

nphilipp:
...what I said on standup that day wasn't just empty promises. It just keeps on growing. :wink:

* Vagrant: fix how the IPA server is provisioned
* Tests: fix one unittest that fails on Python 3.8+, configure Tox to use Python 3.8, too
* Fix import order
* Don't use 3rd party `mock` package (there is `unittest.mock` and we use it in most places already)
* Improve a test which is unnecessarily strict, breaking when unrelated things change


ryanlerch (approved):
LGTM! Just the single conflict to resolve, but other than that. +1

devel/create-test-data.py
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 import textwrap
 
-import python_freeipa
 from faker import Faker
+import python_freeipa
```
abompard, line 4: Are you sure? In [PEP8](https://www.python.org/dev/peps/pep-0008/#id23) they sort imports with `import x` first and `from x import y` second.

In the [Google style guide](https://google.github.io/styleguide/pyguide.html?showone=Imports_formatting#313-imports-formatting) you can see both sortings, and in the [example file](https://github.com/PyCQA/flake8-import-order/blob/master/tests/test_cases/complete_google.py) that flake8-import-order provides the `import x` come before the `from x import y`.

tox.ini
```diff
@@ -74,8 +74,11 @@ commands =
 [flake8]
 show-source = True
 max-line-length = 100
-ignore = E203,E231,W503
 exclude = .git,.tox,dist,*egg
 
+# Configure flake8-import-order
+application-import-names = noggin, noggin_messages
```
abompard, line 80: You need to add `flake8-import-order` to the dev dependencies if you want to use this feature.

abompard (changes requested):

tox.ini
nphilipp: Done.

devel/create-test-data.py
nphilipp: I've changed `import-order-style` to `smarkets` in tox.ini which doesn't require 3rd party packages to be separate from each other, but places `from ... import ...` beneath plain `import ...` statements within their respective blocks.

abompard (approved):

Issue #27, opened 2019-12-23
As a community member (in any capacity), I want spammers to be curtailed ASAP, so that they don’t annoy me.


sfinn85:

relrod:
(This will require some changes to Basset upstream.)

sfinn85:
Sprint planning 3 meeting:

How are they being spammed? More context needed @relrod

abompard:
@relrod, do you know what kind of changes will be required in Basset? How should we integrate with it?

abompard:
Hey folks. This has been opened since the veryt beginning and it still seems unclear what needs to be done. According to [this email thread](https://lists.fedoraproject.org/archives/list/infrastructure@lists.fedoraproject.org/thread/JMWSUIJTJPRHVRLAU22ZEQLOAALPBLDI/#JMWSUIJTJPRHVRLAU22ZEQLOAALPBLDI) Basset needs some work to stay up-to-date with spammer practices and the team considers configuring it to not flag accounts as spam.

If we want to integrate the new AAA system with Basset, we need to know how this can be done, and how we should react to Basset's response.

According to the process described in [our documentation](https://fedora-infra-docs.readthedocs.io/en/latest/sysadmin-guide/sops/basset.html#fas), Basset's integration in the registration process is pretty deep, and we would need to make quite a few changes to replicate it. There would also be security questions to handle, because at the moment Basset itself connects to FAS to set the spam status. With Basset's architecture, we would thus need to write a Basset plugin that would connect to FreeIPA with appropriate permissions and do the equivalent action (moving the user from the stage users category to the active users category).
One of the problems we would face is that FreeIPA does not allow a stage user to change their password, so this spam check can't be done at the end of the regsitration process. One of the ways of doing it would be, for example, to have the following registration process:
1. User registers on the front page
2. Noggin generates a token
3. Instead of sending the token by email for validation, it hands over the token (or rather the full link) to Basset
4. Basset does the spam verification, and if the user is accepted it sends the validation email.
5. User receives the email, clicks on the link and the registration proceeds as before from there.

However, in this process I don't see how we could handle the "manual" spam check status. Noggin has no database, so it would have to be a field set in IPA that admins would go to and somehow check. But what would be the process from there on? The user still hasn't received the validation email and Basset will not know that the admin has approved the user.

[...]

PR #315, opened 2020-07-17
Add Basset support

abompard:
This PR implements the Noggin-side of the workflow described in #27 .

- Users are created with the `spamcheck_awaiting` status.
- Upon registration, a request is sent to Basset to check for spam likelihood. In addition to the user details, this request contains a callback URL and a token that Basset will have to provide with the result
- A webhook endpoint is added to receive Basset's result, which will update the user's status.
- When a user status is not `active`, a landing page is provided to explain to the user that their account is being spamchecked.

### DIFF
--- a/noggin/controller/password.py
+++ b/noggin/controller/password.py
@@ -19,7 +19,7 @@ from noggin.security.ipa import maybe_ipa_session, untouched_ipa_client
 from noggin.utility import messaging, require_self, user_or_404, with_ipa
 from noggin.utility.forms import FormError, handle_form_errors
 from noggin.utility.password_reset import PasswordResetLock
-from noggin.utility.token import PasswordResetToken
+from noggin.utility.token import Audience, make_token, read_token
 from noggin_messages import UserUpdateV1
 
 
@@ -138,7 +138,10 @@ def forgot_password_ask():
                 raise FormError(
                     "username", _("User %(username)s does not exist", username=username)
                 )
-            token = PasswordResetToken.from_user(user).as_string()
+            token = make_token(
+                {"sub": user.username, "lpc": user.last_password_change},
+                audience=Audience.password_reset,
+            )
             # Send the email
             email_context = {"token": token, "username": username}
             email = Message(
@@ -175,11 +178,11 @@ def forgot_password_change():
         flash('No token provided, please request one.', 'warning')
         return redirect(url_for('forgot_password_ask'))
     try:
-        token_obj = PasswordResetToken.from_string(token)
+        token_data = read_token(token, audience=Audience.password_reset)
     except jwt.exceptions.DecodeError:
         flash(_("The token is invalid, please request a new one."), "warning")
         return redirect(url_for('forgot_password_ask'))
-    username = token_obj.username
+    username = token_data["sub"]
     lock = PasswordResetLock(username)
     valid_until = lock.valid_until()
     now = datetime.datetime.now()
@@ -188,7 +191,7 @@ def forgot_password_change():
         flash(_("The token has expired, please request a new one."), "warning")
         return redirect(url_for('forgot_password_ask'))
     user = User(ipa_admin.user_show(username))
-    if not token_obj.validate_last_change(user):
+    if user.last_password_change != token_data["lpc"]:
         lock.delete()
         flash(
             _(
@@ -246,7 +249,10 @@ def forgot_password_change():
             # Oh noes, the token is now invalid since the user's password was changed! Let's
             # re-generate a token so they can keep going.
             user = User(ipa_admin.user_show(username))
-            token = PasswordResetToken.from_user(user).as_string()
+            token = make_token(
+                {"sub": user.username, "lpc": user.last_password_change},
+                audience=Audience.password_reset,
+            )
             form.otp.errors.append(_("Incorrect value."))
         except python_freeipa.exceptions.FreeIPAError as e:
             # If we made it here, we hit something weird not caught above.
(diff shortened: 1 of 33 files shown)


noggin/representation/user.py
```diff
@@ -36,4 +37,4 @@ class User(Representation):
 
     @property
     def name(self):
-        return self.displayname or self.gecos or self.commonname
+        return self.displayname or self.gecos or self.common_name
```
nphilipp: Why this change? Is it to conform with how it's used elsewhere, e.g. in FreeIPA?

noggin/templates/spamcheck.html
```diff
@@ -0,0 +1,51 @@
...
+              {{ _("Your account needs to be approved by an administrator.") }}
+              {{ _("You will be notified by email when the decision has been taken.") }}
+              {{ _("Thank you for your patience.") }}
+              {% elif g.current_user.status_note == "spamcheck_denied" %}
+              {{ _("Your account has been flagged as spam. If it is not correct, please contact the administrators.") }}
+              {{ _("(if it is correct, though, please stop doing that)") }}
+              {% else %}
+              {{ _("Unsupported spam status: %s, please contact the administrators", g.current_user.status_note) }}
+              {% endif %}
+            </p>
```
nphilipp: Is it necessary to wrap the text in Jinja expressions/variables here? If I'm not off, we could change it to something like this to make it easier to read and maintain:

````
{% if g.current_user.status_note == "spamcheck_awaiting" %}
  <h4 class="alert-heading">Your account is being checked</h4>
  <p>
    Your account is being checked for spam likelihood, it should only take a few seconds, please wait...
  </p>
{% elif g.current_user.status_note == "spamcheck_manual" %}
  <h4 class="alert-heading">Your account requires admin activation</h4>
  <p>
    Your account needs to be approved by an administrator.
    You will be notified by email when the decision has been taken.
    Thank you for your patience.
  </p>
{% elif g.current_user.status_note == "spamcheck_denied" %}
￼  <h4 class="alert-heading">Your account is blocked</h4>
  <p>
    Your account has been flagged as spam. If it is not correct, please contact the administrators.
    (If it is correct, though, please stop doing that.)
  </p>
{% else %}
  <h4 class="alert-heading">Something went wrong</h4>
  <p>
    Unsupported spam status: {{ g.current_user.status_note }}, please contact the administrators.
  </p>
{% endif %}
````

Also, we probably should tell people how to contact administrators -- they're new to the project and won't be aware of the various channels.

File: noggin/controller/password.py [hunk 3]
```diff
@@ -138,7 +138,10 @@ def forgot_password_ask():
                 raise FormError(
                     "username", _("User %(username)s does not exist", username=username)
                 )
-            token = PasswordResetToken.from_user(user).as_string()
+            token = make_token(
+                {"sub": user.username, "lpc": user.last_password_change},
+                audience="pwdrst",
```
nphilipp: Am I right assuming that the valid values of `audience` are defined in Basset? I had to think hard what `pwdrst` means (and I'm at an advantage, I see what this was before :wink:).

nphilipp (changes requested):

noggin/representation/user.py
abompard: Hmm not really, this change isn't really necessary. I don't know what's best. I'll revert it since after all we already have `displayname` without the underscore.

noggin/templates/spamcheck.html
abompard: That's indeed a bit easier to read, I'll switch to that.

noggin/controller/password.py
abompard: No, audience values are internal, it just needs to be the same on token creation and token parsing side, to avoid token re-use. We can use whatever we want but using too long a string has disadvantages because it makes the token longer, and it ends up in the URL so there's a size limit. I could use an enum if you think it makes it clearer.

abompard:
Thanks for the review Nils, I've updated the PR

noggin/controller/password.py
nphilipp: Maybe it's just my brain taking the wrong turn when coming across anything `pwd...`, maybe we could use `passreset` (which is only one character more)?

abompard: I switched to an enum, it should make things clearer :-)

noggin/templates/spamcheck.html
```diff
@@ -0,0 +1,58 @@
...
+    <div class="container section">
+      <div class="row">
+        <div class="col-md-7">
+          {{ front_page_blurb() }}
+        </div>
+        <div class="col">
+          <div class="alert alert-{% if g.current_user.status_note == 'spamcheck_denied' %}warning{% else %}info{% endif %}" role="alert">
+            {% if g.current_user.status_note == "spamcheck_awaiting" %}
+              <h4 class="alert-heading">
+                {{ _("Your account is being checked") }}
```
nphilipp, line 16: Again, why does it have to be wrapped in a Jinja expression like this? I'm used to messages being extracted for translation from templates and replaced with localized versions when rendered, just as they are.

nphilipp (changes requested):

noggin/templates/spamcheck.html
nphilipp: To clarify my question, why is this...:

````
{{ _("Your account is being checked") }}
````

...not this?

````
Your account is being checked
````

abompard: Not sure what you mean, do you have an example? I'm following [this doc](https://jinja.palletsprojects.com/en/2.11.x/extensions/#new-style-gettext).

pypingou: @nphilipp for translations :)

nphilipp: I've created a PR covering "we should do l10n in templates differently" here: https://github.com/fedora-infra/noggin/issues/317

nphilipp (approved):
Looks good to me!

PR #322, opened 2020-08-03
New workflow for spam checking

abompard:
The spam checking is requested before the user had a chance to confirm their email address. This way we don't need to lock the account, and we workaround the fact that IPA does not give us the reason why an account is locked.

It's also cleaner to have all potential spam accounts stay in the stageuser state, it's easier to round them up and delete them after a while, should we choose to do so.


nphilipp (approved):
Looks good to me!

PR #326, opened 2020-08-06
Dynamically create the app object

abompard:
Instead of instanciating the app as a global object, use the [application factory](https://flask.palletsprojects.com/en/1.1.x/patterns/appfactories/) pattern as recommended by Flask.

This means we have a root blueprint and register the controllers with it rather than the main app, and as a consequence we need to prefix all endpoints with a dot, which is what causes most of the changes in the templates.

Nice side effect: we don't need to pass the configuration environment variable just to run the unit tests anymore, which make them more VSCode-friendly. We also don't have to set that environment variable to build the docs or run `towncrier` either.

Also, reorganize some of the code to have a clearer separation of module topics.

Note: the last commit is only a rebuild of the cassettes.


noggin/controller/__init__.py
```diff
@@ -0,0 +1,21 @@
+from flask import Blueprint, g, render_template, session
+
+from noggin.utility.templates import gravatar
+
+
+blueprint = Blueprint("root", __name__)
```
nphilipp, line 6: I think `blueprint` is a little too generic (could be `root` to match the name of the blueprint), but this shouldn't block merging.

noggin/security/ipa_admin.py
```diff
@@ -70,7 +79,7 @@ def wrapper(*args, **kwargs):
 
     def __getattr__(self, name):
         wrapped_methods = list(self.__WRAPPED_METHODS)
-        if self.__app.config['TESTING']:  # pragma: no cover
+        if current_app.config.get('TESTING', False):  # pragma: no cover
```
nphilipp, line 82: Putting in `False` as the default is superfluous, as `None` (the "default default" :wink:) would work fine. As well, shouldn't block.

noggin/tests/unit/test_middleware.py
```diff
@@ -12,3 +14,10 @@ def test_ipa_error(client, mocker):
     page = BeautifulSoup(result.data, 'html.parser')
     assert page.title
     assert page.title.string == 'IPA Error - noggin'
+
+
+def test_flask_ext(mocker):
+    init_app = mocker.patch.object(IPAErrorHandler, "init_app")
+    dummy_app = object()
+    IPAErrorHandler(dummy_app)
+    init_app.assert_called_once_with(dummy_app)
```
nphilipp, line 23: This looks like it's the same as in `noggin/tests/unit/security/test_ipa_admin.py` above. Is there a reason for that, or are they slightly different?

nphilipp (approved):

noggin/tests/unit/test_middleware.py
nphilipp: Ahh, it's mocking in a different object.

PR #329, opened 2020-08-07
Make first name and last name optional

mscherer:
Asking names when this is not needed go against GDPR spirit
(Art 5.c: "data minimization"), and since FAS do not ask that, it
is hard to argue we need the information.

This is also a problems for people having more than 1 name,
or people having just 1
(see https://en.wikipedia.org/wiki/Mononymous_person#Modern_times ).

There is also people who do not want to give their name for whatever reasons,
and who could do that before with FAS.

Finally, since the information is requested by default without opt-out, someone
who would change last name or first name will be forced to change it in FAS, thus adding
to the burden of name change. In the western world, that mean mostly
woman and/or trans people. So not having opt-out do not get in a direction
of simplifying their life, which seems to not go in favor of diversity.

Signed-off-by: M. S <misc@redhat.com>

### DIFF
--- a/noggin/form/edit_user.py
+++ b/noggin/form/edit_user.py
@@ -21,12 +21,12 @@ from .base import BaseForm, CSVListField
 class UserSettingsProfileForm(BaseForm):
     firstname = StringField(
         _('First Name'),
-        validators=[DataRequired(message=_('First name must not be empty'))],
+        validators=[Optional()],
     )
 
     lastname = StringField(
         _('Last Name'),
-        validators=[DataRequired(message=_('Last name must not be empty'))],
+        validators=[Optional()],
     )
 
     mail = EmailField(
--- a/noggin/form/register_user.py
+++ b/noggin/form/register_user.py
@@ -4,7 +4,7 @@ from wtforms.fields.html5 import EmailField
 from wtforms.validators import DataRequired, EqualTo, Length
 
 from noggin import app
-from noggin.form.validators import Email
+from noggin.form.validators import Email, Optional
 
 from .base import BaseForm, ModestForm, strip, SubmitButtonField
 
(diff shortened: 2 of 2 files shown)


mscherer:
While I know there is problem in more than noggin, and this PR wouldn't be accepted due to that, I opened it to start the discussion after mentioning the topic on #fedora-diversity irc chan after the presentaton at Nest with Fedora.

We can't really say "yeah, diversity" and at the same time, making things harder for some non western folks, women and trans people.

ryanlerch:
Are we able to make this one an issue here in Noggin? AIUI, the need to have a firstname and a lastname is an IPA requirement, not a noggin one, and the code being changed with the PR here in noggin is nothing more than form validation to ensure we send valid data to IPA.

nphilipp:
> Are we able to make this one an issue here in Noggin? AIUI, the need to have a firstname and a lastname is an IPA requirement, not a noggin one, and the code being changed with the PR here in noggin is nothing more than form validation to ensure we send valid data to IPA.

@ryanlerch I think that's what @mscherer was getting at with the previous comment.

Right now, after fixing the conflicts and importing `Optional` from the right place (`wtforms.validators`), the test suite runs fine, but: I'd like this PR to get tests which cover the use cases (with and without names), with which we can verify that Noggin + FreeIPA together are able to handle this.

As to the discussion in general, if FAS previously let people register without divulging their name, then Noggin/FreeIPA shouldn't start to do so, for all the reasons outlined in the initial comment. You wouldn't even have to bring the GDPR into it, in that light alone I think it could be said that Fedora has a reasonable interest in knowing who its contributors are because their activities might have legal ramifications for the project. But that's a totally different can of worms I don't want to open up here. :wink:

PR #340, opened 2020-08-13
Proposed changes to #326 that didn't warrant blocking

nphilipp:


ryanlerch:
@nphilipp is this still something that we want?

ryanlerch:
@nphilipp is this still something you think we need?

PR #346, opened 2020-08-19
Remove some wrapped methods that ipa-admin never uses

abompard:
Those methods are never actually called in the code.


nphilipp:
Looks good to me!

nphilipp (approved):

PR #367, opened 2020-09-23
Set a BASE_URL variable in the master template to be used in search.js

pypingou:
Basically this variable allows to specify the full URL where noggin
is deployed. Then in search.js we are able to build the full URL
linking to where the user/group are.

The current behavior assumed that noggin was running at the root of
the web-server (ie: /) while this is not necessarily true (it could
run at say /accounts/).

This commit should fix that.

Signed-off-by: Pierre-Yves Chibon <pingou@pingoured.fr>


abompard:
This was fixed in #362, it's pending but I'll merge it anyway so we don't duplicate work in the future again, sorry

abompard:
Our reviewing throughput has been slower than usual these days, which is why the other PR has been pending for a while.

PR #368, opened 2020-09-25
Install python3-devel RPM package

areguera:
- Otherwise, the "poetry install" command fails.


areguera:
This is duplicated. #350 already considers the python3-devel RPM package inclusion.

PR #389, opened 2020-10-23
Fix returned search URL in mounted app

nphilipp:
Additionally, test the endpoint for both cases (application root set & unset), and add a context manager to easily tweak the Flask application configuration for that purpose.


nphilipp:
Adapting the tests for this took me a while.

My first approach was a combination of enabling parametrizing the `app_config` fixture with decorators for the test functions/methods, but that turned out to require recreating all the VCR cassettes (because parametrizing the fixture changes the signature of tests being run, yay). And for some reason that triggered all kinds of test breaking behavior, so I ditched it :wink:.

The way it's tested now is much less intrusive, the new `tweak_app_config` context manager (in `nogging.tests.unit`) accepts a dictionary with configuration keys, values to replace or add to the configuration. Upon exiting the context, it restores old values and deletes keys that weren't present before from the configuration dictionary in the app object.

abompard:
I don't think this is what the `APPLICATION_ROOT` variable is for. According to the [documentation](https://flask.palletsprojects.com/en/1.1.x/config/#APPLICATION_ROOT), it's only used to build URLs outside of the context of a request, which is not our case here.
Actually, I don't even think that this change is necessary. The internal URL should be sufficient, the external URL (which adds protocol and hostname) don't add the mount path, it should already be there in the internal URL.
When I look at the staging deployment it seems to not return a `url` at all so I think the problem is that the staging deployment does not have commit 3e7cbdb5 yet (which adds the `url` value). I'll update it and see if it's fixed.

abompard:
Yep, that was it. Sorry...

nphilipp:
> I don't think this is what the `APPLICATION_ROOT` variable is for. According to the [documentation](https://flask.palletsprojects.com/en/1.1.x/config/#APPLICATION_ROOT), it's only used to build URLs outside of the context of a request, which is not our case here.

Ah, it seems to be `SCRIPT_NAME` which we set as an environment variable in `roles/openshift-apps/noggin/templates/deploymentconfig.yml`. I wonder why we set `APPLICATION_ROOT` in `roles/openshift-apps/noggin/templates/noggin.cfg.py`, though.

abompard:
Not sure, it would be relevant if we were running command line scripts or cron jobs, but I don't think we do.

Issue #332, opened 2020-08-11
[secaudit-blocking] No installation documentation


puiterwijk:
Part of secaudit #316, blocking.

The [Installation documentation](https://github.com/fedora-infra/noggin/blob/0e3be29de02a1ba7aaf247493c5adf7d08e5f64b/docs/installation.rst) is absent, and doesn't contain any useful information in how to set up a production-grade setup of noggin.

ryanlerch:
@puiterwijk we will get working on this ASAP. 

Note too, (while not documentation, i know), there is the vagrant setup that sets up a freeipa server, installs the freeipa-fas plugin to freeipa, then installs noggin.

Note too there is an in-review PR here (https://github.com/fedora-infra/noggin/pull/326) that makes the noggin flask app behave a little better.

puiterwijk:
@ryanlerch Right. But note that if you decide the Vagrant setup is the official way of deploying, I'll have many complaints about insecure deployment practices 😀. That's why I want to see how you tell people to actually deploy it.

nphilipp:
@puiterwijk I don't think we'll document using Vagrant to deploy in production, with or without your comment :wink:. Off the top of your head, are there any other gotchas? If we can avoid embarrassing ourselves upfront, I'm all for it.

puiterwijk:
Some of the things you really want to point out:
- Don't use `flask run`, but instead use a serious HTTP server, and explain how to do so (apache/nginx with possible gunicorn behind it)
- Do not use the global `admin` user: instead, I'd *strongly* recommend a separate noggin user for auditing/permission purposes (as part of the docs would then be "What are the minimum required permissions to grant")
- Make very sure to change `SECRET` and `FERNET_SECRET` (#334)

And other things like those

Conan-Kudo:
@nphilipp @puiterwijk Probably a good starting point for installation documentation would be the haphazard one I wrote for getting the system up and running for openSUSE infrastructure on COPR: https://copr.fedorainfracloud.org/coprs/ngompa/fedora-aaa/

Cliftonz:
Any updates on this?

abompard:
[...]

PR #391, opened 2020-10-27
Noggin installation docs

davidkirwan:
References issue: #332

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,5 +1,238 @@
 ============
 Installation
 ============
+Installation documentation for Noggin.
 
-.. note:: **TODO**: Cover end-user installation here.
+
+============
+Requirements
+============
+Noggin requires the following to operate:
+
+1. Access to a FreeIPA/Red Hat Identity Management installation
+2. Access to an Openshift cluster to run the Noggin container
+
+
+===================================
+Install Red Hat Identity Management
+===================================
+For more information see the following documents regarding installation
+
+1. .. _Docs: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/installing_identity_management/preparing-the-system-for-ipa-server-installation_installing-identity-management
+2. .. _Playbook install docs: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/installing_identity_management/installing-an-identity-management-replica-using-an-ansible-playbook_installing-identity-management
+3. .. _Ansible FreeIPA module docs: https://github.com/freeipa/ansible-freeipa
+
+
+Install ansible and ansible-freeipa rpms
+
+::
+    yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+    ARCH=$( /bin/arch )
+    subscription-manager repos --enable "codeready-builder-for-rhel-8-${ARCH}-rpms"
+    yum install ansible
+    yum install ansible-freeipa
+    yum install freeipa-fas
+
+
+The location where the ansible roles are installed:
+
+::
+    [root@server]# ls -1 /usr/share/ansible/roles/
+    ipaclient
+    ipareplica
+    ipaserver
+
+
+The location where the documentation is once installed:
+
+::
+    [root@server]# ls -1 /usr/share/doc/ansible-freeipa/
+    playbooks
+    README-client.md
+    README.md
+    README-replica.md
+    README-server.md
+    README-topology.md
+
+
+The location where the ansible playbooks are once installed:
+
+::
+    [root@server]# ls -1 /usr/share/doc/ansible-freeipa/playbooks/
+    install-client.yml
+    install-cluster.yml
+    install-replica.yml
+    install-server.yml
+    uninstall-client.yml
+    uninstall-cluster.yml
+    uninstall-replica.yml
+    uninstall-server.yml
+
+
+Sample Inventory:
+
+::
+    [root@server]# cat inventory
+    [ipaservers]
+    ipa01.stg.iad2.fedoraproject.org
+
+    [ipareplicas]
+    ipa02.stg.iad2.fedoraproject.org
+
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -1,5 +1,145 @@
...
+    [os_stg:children]
+    os_nodes_stg
+    os_masters_stg
+    os_control_stg
+
+
+Run the installation playbook:
+
+::
+    [root@server]# ansible-playbook -i inventory/inventory playbooks/openshift-apps/noggin.yml
```
puiterwijk: This is definitely going to need some further configuration with secrets etcetera, because clones of `ansible.git` won't have that outside of the Fedora batcave.

docs/installation.rst
```diff
@@ -1,5 +1,145 @@
...
+    ---
+    - name: Playbook to configure IPA server
+      hosts: ipaserver
+      become: true
+
+      roles:
+      - role: ipaserver
+        state: present
+
+    [root@server]# ansible-playbook -i inventory install_server.yaml
```
puiterwijk, line 107: The installation instructions of FreeIPA didn't have to be in there, but nice that they are. However, I'm missing instructions on creating service accounts (with minimal privileges), and for example freeipa-fas.

puiterwijk (changes requested):
I am missing a couple of things in here, and I would not consider this enough for #332. I'm missing for example instructions on installing freeipa-fas, on setting up service accounts, on configuring noggin, etcetera.
Also, I'm not sure that "use the Fedora Infra ansible repo" is a very maintainable way of doing this: the Fedora Infra ansible repo will evolve significantly over time, and you'd have to make sure that those changes get release branches (when config variables etc change), or other things.

docs/installation.rst
```diff
@@ -1,5 +1,145 @@
...
+For more information see the following documents regarding installation
+
+1. .. _Docs: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/installing_identity_management/preparing-the-system-for-ipa-server-installation_installing-identity-management
+2. .. _Playbook install docs: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/installing_identity_management/installing-an-identity-management-replica-using-an-ansible-playbook_installing-identity-management
+
+
+Install ansible and ansible-freeipa rpms
+
+::
+    subscription-manager repos --enable ansible-2.8-for-rhel-8-x86_64-rpms
```
Conan-Kudo: Please do not use RHEL as the sole reference documentation for how to install and deploy this. For one, stuff like `subscription-manager` and access to various RH product subscriptions are not going to be available by most people. It's fine to have RHEL specific notes, but please write the installation process using CentOS and Fedora instead, since that's what pretty much everyone has access too.

Conan-Kudo (changes requested):

Conan-Kudo:
@puiterwijk I can start landing the Noggin stack components in Fedora and EPEL as appropriate to help with this...

docs/installation.rst
davidkirwan: Yes I agree, RHEL installation docs should not be the sole reference but it will be the first documented reference. We've made a decision to first document how we plant to install Noggin in Fedora.

docs/installation.rst
```diff
@@ -1,5 +1,170 @@
...
+
+    # These variables will be used when env == 'dev' or set to any value other than 'production'
+    # noggin_stg_admin_password: ""
+    # noggin_stg_fernet_secret: ""
+    # noggin_stg_session_secret: ""
+    # noggin_stg_github_secret : "" # Openshift github webhook, not needed
+    # env: "staging"
+    # env_suffix: ".stg"
+
+    noggin_theme: ""                                                           #
```
davidkirwan: @abompard I'm not sure what goes here in the `noggin_theme` variable.

docs/installation.rst
Conan-Kudo: Then don't use ansible-for-rhel and use ansible from EPEL, because then you can minimize the delta between RHEL and CentOS.

docs/installation.rst
davidkirwan: wait nm found it, https://github.com/fedora-infra/noggin/blob/dev/noggin.cfg.default#L13-L14

So "default" is enough.

docs/installation.rst
davidkirwan: I think I've addressed this part of the requested changes, I've added the ansible variables which are used within the playbook.

davidkirwan:
> 
> 
> @puiterwijk I can start landing the Noggin stack components in Fedora and EPEL as appropriate to help with this...

@Conan-Kudo  I'd love to sit in and shadow you doing that, I've only a vague understanding of whats required!

docs/installation.rst
davidkirwan: I think I've addressed this @Conan-Kudo but I'm not sure, mind taking another look when you get a chance thanks.

Conan-Kudo:
@davidkirwan Sure, let me know when. 👍🏾

docs/installation.rst
```diff
@@ -1,5 +1,197 @@
...
+
+::
+    yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+    ARCH=$( /bin/arch )
+    subscription-manager repos --enable "codeready-builder-for-rhel-8-${ARCH}-rpms"
+    yum install ansible
+    yum install ansible-freeipa
+    yum install yum-plugin-copr
+    yum copr enable ngompa/fedora-aaa
+    yum install freeipa-fas
```
Conan-Kudo, line 34: `freeipa-fas` has been submitted to EPEL 8: https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2020-3f20306084

docs/installation.rst
davidkirwan: @puiterwijk I think I've addressed this, can you take another look please!

PR #397, opened 2020-11-12
Pagination: fix links when mounted in a subdirectory

### LINKED ISSUE #396: Broken pagination link on the group page
When accessing https://admin.stg.fedoraproject.org/accounts/groups/ if you scroll down to the bottom and try to use the pagination links, it sends you to: https://admin.stg.fedoraproject.org/groups/?page_size=30&page_number=2 which doesn't work as it is missing the ``/accounts/`` part of the URL.

abompard:
Fixes: #396

### DIFF
--- a/noggin/utility/pagination.py
+++ b/noggin/utility/pagination.py
@@ -30,7 +30,7 @@ class PagedResult:
         qs = dict(request.args)
         qs.update({"page_size": self.page_size, "page_number": page_number})
         qs = "&".join(f"{k}={v}" for k, v in qs.items())
-        return f"{request.path}?{qs}"
+        return f"{request.script_root}{request.path}?{qs}"
 
     def __repr__(self):
         return f"<PagedResult items=[{len(self.items)} items] page={self.page_number}>"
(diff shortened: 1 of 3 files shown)


noggin/utility/pagination.py
```diff
@@ -30,7 +30,7 @@ def page_url(self, page_number):
         qs = dict(request.args)
         qs.update({"page_size": self.page_size, "page_number": page_number})
         qs = "&".join(f"{k}={v}" for k, v in qs.items())
-        return f"{request.path}?{qs}"
+        return f"{request.script_root}{request.path}?{qs}"
```
pypingou, line 33: I was looking into this and I tried it with ``request.base_url`` anything against using the full path directly?

abompard: It would work too, it's just that `base_url` would give the full URL including scheme and hostname, and we don't need that. But both would work really. I usually avoid full URLs for internal links because I've seen it fail too many times when there's a proxy that does SSL termination in front of the app, but in our case that possibility of failure is already taken care of, so both ways are equivalent I think.

PR #413, opened 2020-12-09
Don't load from CDN

### LINKED ISSUE #337: [secaudit-blocking] Multiple remote resources used not on approved origins
Part of secaudit #316, blocking.

The [Fedora Infrastructure Application Security Policy](https://docs.pagure.org/infra-docs/dev-guide/security_policy.html#security-headers) lists the origins that are allowed for the CSP header. Any resources loaded from origins not listed there will be rejected by the browser, and as such are not allowed to be used.
Noggin currently refers to many external resources, for example [`https://stackpath.bootstrapcdn.com/`](https://github.com/fedora-infra/noggin/blob/0e3be29de02a1ba7aaf247493c5adf7d08e5f64b/noggin/themes/default/templates/main.html#L5), [`https://code.jquery.com/`](https://github.com/fedora-infra/noggin/blob/0e3be29de02a1ba7aaf247493c5adf7d08e5f64b/noggin/themes/default/templates/main.html#L67) and [`https://cdnjs.cloudflare.com/`'](https://github.com/fedora-infra/noggin/blob/0e3be29de02a1ba7aaf247493c5adf7d08e5f64b/noggin/themes/default/templates/main.html#L68).

abompard:
It is against Fedora's security policy.

Fixes: #337


Conan-Kudo (approved):

Conan-Kudo:
@abompard Any JavaScript we bundle also needs its non-minified sources included too, even if we're not using it. We do this in Pagure, for example: https://pagure.io/pagure/blob/master/f/pagure/static/vendor

ryanlerch (approved):

PR #426, opened 2021-01-13
add the ability to set / view pronouns

### LINKED ISSUE #220: As a registered user, I want to record the pronoun I use so people use the right one when discussing with me
Close to #105, but being able to show what pronouns people use can be a big deal for inclusivity. 

Also https://github.com/fedora-infra/noggin-old/issues/9 

From a technical point of view, there isn't anything complex (just store the information), but the real problem is to decide on limitation and format to avoid abuse.

There is a few pitfalls to avoid, and on top of my head:

- some people misunderstand the purpose, most likely due to mistranslation and/or lack of awareness. For example, I heard several time the story of austrian/german people using that kind of field to add their title "dr", because that's quite a big deal there. In the same way, if we ever get a lord/lady in UK in the community, maybe they would use that. I would recommand having a title entry to avoid that, but that would be a separate bug.

- there is a tension between "being very inclusive" (using a free form field, since people use new pronouns all the time), and "avoiding folks putting jokes there".  For example, at RH (where did deploy that on our internal directory since 2 or 3 years), I did some stats and we had 500 to 1000 people using the system at first, and around 20 who didn't really used it in good faith.  

- some people might not want to use that, some people might want to use more than 1 pronoun. So being able to opt out, and being able to select more than 1 seems important. 

based on research, there is 2 way.

The first one is a free form field. This is what we have inside RH, and while that's compact, this requires some moderation. On top of that, since there is no mandated format, it is not trivial to do stats, nor easy to signal people use more than one. For example, folks did use "she/her", "she, her", "she her", but also "she, they". On the plus side, it didn't limit what can be used, which was better for inclusivity.

On the other end of the spectrum, there is what Google did deploy, which is a set of checkbox, so people can choose one or more in a set of 10 different choices. I wasn't able to get more information on that. Based on my stats inside RH, I would say that he/she/they did cover almost everybody, except 2 or 3 persons. This do not requires moderation, but is not compact.

One middle ground would be checkbox for commons (he/she/they) + a free form field.

I also spent some time looking for guidance on the topic, and did found a few ressources, if that help:
https://queerlyrepresent.me/resources/articles/gender-accessibility
https://uxdesign.cc/designing-forms-for-gender-diversity-and-inclusion-d8194cf1f51
https://www.brynmawr.edu/sites/default/files/asking-for-name-and-pronouns.pdf

Finally, there is also the question on whether this information should be used as a proxy for grammatical gender or not in interface, which do bring the question on how it should be stored (and so used. 

While that's tempting, I would suggest "no", given the relative complexity of grammatical gender around the world, and the fact that some languages are more gendered than english, which would bring all kind of interesting problems.

ryanlerch:
Adds an additional field to the user profile, allowing
a user to set their pronouns.

Fixes: #220
Related: fedora-infra/freeipa-fas#134

### DIFF
--- a/noggin/form/edit_user.py
+++ b/noggin/form/edit_user.py
@@ -76,6 +76,8 @@ class UserSettingsProfileForm(BaseForm):
         validators=[Optional()],
     )
 
+    pronouns = StringField(_('Pronouns'), validators=[Optional()],)
+
 
 class UserSettingsKeysForm(BaseForm):
     sshpubkeys = FieldList(
--- a/noggin/controller/user.py
+++ b/noggin/controller/user.py
@@ -131,6 +131,7 @@ def user_settings_profile(ipa, username):
                 'fasrhbzemail': form.rhbz_mail.data,
                 'faswebsiteurl': form.website_url.data,
                 'fasisprivate': form.is_private.data,
+                'faspronoun': form.pronouns.data,
             },
             ".user_settings_profile",
         )
--- a/noggin/representation/user.py
+++ b/noggin/representation/user.py
@@ -26,6 +26,7 @@ class User(Representation):
         "status_note": "fasstatusnote",
         "creation_time": "fascreationtime",
         "is_private": "fasisprivate",
+        "pronouns": "faspronoun",
     }
     attr_types = {
         "sshpubkeys": "list",
--- a/noggin/templates/user-settings-profile.html
+++ b/noggin/templates/user-settings-profile.html
@@ -26,6 +26,7 @@
         <div class="form-group col">{{ macros.with_errors(form.firstname) }}</div>
         <div class="form-group col">{{ macros.with_errors(form.lastname) }}</div>
       </div>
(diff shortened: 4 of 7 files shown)


noggin/form/edit_user.py
```diff
@@ -76,6 +76,8 @@ class UserSettingsProfileForm(BaseForm):
         validators=[Optional()],
     )
 
+    pronouns = StringField(_('Pronouns'), validators=[Optional()],)
```
abompard, line 79: If we provide the full flexibility of a string field here, I'm worried that we'll end up with vastly different values that are actually pretty much the same ("she", "she/her", "woman", etc...).
At first I think we should provide a select field with "she / her", "he / him", "they / them" and "other", to limit possibilities.
Then we could find a way for a free-form field to appear when "other" is selected. I've tried mocking that up but it's taking me too long (jquery...) so I think we can start with the simple select field. What do you think?

ryanlerch: updated to provide some default pronoun sets

abompard (approved):
Looks great, thanks!

PR #435, opened 2021-02-02
Add an optional note to the user settings OTP page

### LINKED ISSUE #433: [RFE] Add warning/info text box in noggin for OTP
As found out during some tests, and discussed in #fedora-aaa, it seems that enabling OTP enforces 2FA everywhere.
So a user who has enabled it would find himself in troubles for simple `kinit`.
While we can have another card/spike for what needs to be done to have `pkinit` working externally (it doesn't actually, but it's been evaluated), this card is more to just had warning/info box in /user/<user>/settings/otp/ that would explain the `kinit` change and how to get kerberos ticket (if people want to build in koji, query fasjson, etc)

ryanlerch:
Add an optional note to the user settings OTP page that is defined in
the main.html in the themes templates to display a note at the top of
the OTP page.

This is used in Fedora Accounts to point users at the documentation for
getting a kerberos ticket when OTP is enabled.

Resolves: #433 

### DIFF
--- a/noggin/templates/user-settings-otp.html
+++ b/noggin/templates/user-settings-otp.html
@@ -93,6 +93,7 @@
       </div>
     </div>
   </div>
+  {{ otp_notice() if otp_notice is defined }}
 {% else %}
   <div class="list-group-item text-center bg-light text-muted font-weight-bold">
     <div>{{ _("You have no OTP tokens") }}</div>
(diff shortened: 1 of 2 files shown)


arrfab:
Just curious, wouldn't that also require same note in all other themes ?

markobrien1:
lgtm

PR #444, opened 2021-02-08
Update search with loading animations

ryanlerch:
Previously, there were no loading animations for the search box, and due
to the fact we are doing two seperate calls (for users and groups),
these could return at different times. So this commit adds sepearte
spinner animations for the two results classes.

Additionally, this moves the string for "users" and "groups" that is
used to label the groups in the search dropdown into the templates as
variables, so they can be translated.

### DIFF
--- a/noggin/static/js/search.js
+++ b/noggin/static/js/search.js
@@ -24,14 +24,18 @@ $("#search")
       display: "uid",
       source: bloodhound_users,
       templates: {
-        header: "<h5>Users</h5>",
+        header: "<h5 class='mb-0'>"+USERS_SEARCH_LABEL+"</h5>",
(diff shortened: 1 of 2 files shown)


noggin/static/js/search.js
```diff
@@ -19,19 +19,23 @@ var bloodhound_groups = new Bloodhound({
 $("#search")
   .typeahead(
     { highlight: true },
-    {
+    {
```
nphilipp: Please don't introduce trailing white space. :wink:

nphilipp (changes requested):
Other than the white space issue, this looks good. Let me see if I can push the fix and merge.

noggin/static/js/search.js
nphilipp: Fixed.

nphilipp (approved):

PR #497, opened 2021-02-12
Be able to list more than 100 groups, plus deps fixes

### LINKED ISSUE #438: The groups list is limited to 100 entries
By default the IPA API does not return more than 100 entries, therefore the groups page only displays the 100 first entries. This can be overridden by setting `sizelimit=0` in the query.

nphilipp:


StephenCoady (approved):

abompard (approved):
We could also set the `sizelimit` directly in the `paginated_find` function because I think we'll want to raise it to the max each time we expect we'll need pagination.
But it's fine as is, we can always refactor later when we make more use of `paginated_find`.

nphilipp:
> We could also set the `sizelimit` directly in the `paginated_find` function because I think we'll want to raise it to the max each time we expect we'll need pagination.
> But it's fine as is, we can always refactor later when we make more use of `paginated_find`.

I went and set `sizelimit=0` in `paginated_find` directly.

PR #500, opened 2021-02-16
OTP: Use default token algorithm (SHA1)

### LINKED ISSUE #441: TOTP tokens could be more compatible with the standard
From Crys on IRC:

Google authenticator does not support SHA-256, only SHA-1 HMAC.
TOTP standard is HMAC-SHA1 with 6 digits. FreeIPA supports other HMAC algorithms and more digits.
That's the culprit: `o_ipatokenotpalgorithm='sha512'`
The default value is fine, it's SHA-1, which is not a security problem. TOTP uses truncated HMAC.

nphilipp:
Previously, we set this to SHA512 but not all authenticator apps are
compatible with that.

Signed-off-by: Nils Philippsen <nils@redhat.com>

### DIFF
--- a/noggin/controller/user.py
+++ b/noggin/controller/user.py
@@ -184,9 +184,7 @@ def user_settings_otp(ipa, username):
         try:
             maybe_ipa_login(current_app, session, username, addotpform.password.data)
             result = ipa.otptoken_add(
-                o_ipatokenowner=username,
-                o_ipatokenotpalgorithm='sha512',
-                o_description=addotpform.description.data,
+                o_ipatokenowner=username, o_description=addotpform.description.data,
             )['result']
 
             uri = urlparse(result['uri'])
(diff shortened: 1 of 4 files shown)


ryanlerch:
Unit-tests are passing, but the tests are currently failing on the format (i.e. black) check.

```
> poetry run black --check .
would reformat /home/rlerch/Source/noggin/noggin/tests/unit/conftest.py
would reformat /home/rlerch/Source/noggin/noggin/controller/user.py
would reformat /home/rlerch/Source/noggin/noggin/tests/unit/controller/test_user_otp.py
Oh no! 💥 💔 💥
3 files would be reformatted, 87 files would be left unchanged.
```

abompard:
I fixed some Black issues in the dev branch, could you rebase please?

nphilipp:
Not sure what happened here, but black complains about things elsewhere. I've prepended a commit to re-black things.

abompard (approved):

PR #507, opened 2021-02-18
Added links to the Fedora Accounts / CentOS Accounts docs

ryanlerch:
Also added the otp note that points to the docs if users
have OTP setup with the details aboout OTP and kinit


nphilipp:
LGTM!

PR #524, opened 2021-03-06
Lower minimum age for registration to 13 to match FAS

### LINKED ISSUE #518: As a project, we want to be compliant with COPPA in the US and disallow creation of accounts for children under 13 years old
https://www.ftc.gov/enforcement/rules/rulemaking-regulatory-reform-proceedings/childrens-online-privacy-protection-rule

The current new account creation workflow in FAS involves checking off that you are at least 13 years of age when creating an account. Registration in the current stg server appears to be turned off so I'm not sure if this verification is available in new account creation now, so this is just a note to make sure it is.

Conan-Kudo:
The United States Children's Online Privacy Protection Act (U.S COPPA)
states the minimum age is 13. Changing this number has other undesirable
consequences, so we should not unless there's an actual reason to do so.

Fixes: 47b8b42ffb68f3a954c31df9c840f44ad8ab640e
Fixes: #518

### DIFF
--- a/noggin/form/register_user.py
+++ b/noggin/form/register_user.py
@@ -37,10 +37,10 @@ class RegisterUserForm(ModestForm):
     )
 
     underage = BooleanField(
-        _('I am over 16 years old'),
+        _('I am over 13 years old'),
         validators=[
             DataRequired(
-                message=_("You must be over 16 years old to create an account")
+                message=_("You must be over 13 years old to create an account")
             )
         ],
     )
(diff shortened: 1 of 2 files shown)


ryanlerch:
@abompard do you remember the story behind this? did legal say 16, that is why we do 16 in noggin?

just not sure if we still need to do this.

ryanlerch:
this one has been open for a while, do we have any guidance on legal if we can even change this now?

going to close, feel free to re-open if and when we get guidance from legal

PR #528, opened 2021-03-08
Use the correct address for OpenID endpoints for openSUSE

Conan-Kudo:
openSUSE set up Ipsilon in a different configuration, where the
OpenID endpoint isn't `<username>.id.opensuse.org`, but instead a
regular URL path where the username is at the end of the URL.

### DIFF
--- a/noggin/templates/user-settings-profile.html
+++ b/noggin/templates/user-settings-profile.html
@@ -12,7 +12,7 @@
     {% set buttontext = _('Change Avatar')%}
     {%- if OPENID is defined %}
     <form method="GET" action="https://www.libravatar.org/openid/login/">
-      <input type="hidden" name="openid_identifier" value="{{ 'https://'+user.username+OPENID }}">
+      <input type="hidden" name="openid_identifier" value="{{ OPENID % (user.username) }}">
       <input name="change-avatar" type="submit" value="{{ buttontext }}" class="btn btn-sm btn-outline-primary">
     </form>
     {%- else %}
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
cc: @hellcp

hellcp:
This won't really work anyway, since our openid isn't setup for `$username.id.opensuse.org`, but `www.opensuse.org/openid/user/$username` iirc. I implemented the openid code here before to support such a usecase, but that didn't stay after some changes

Conan-Kudo:
@hellcp Any chance we can get this corrected then?

hellcp:
Just copy aadd9f03559597b44b572dd71757fbf87bdc51ba

Conan-Kudo:
@hellcp How does this look now?

hellcp:
Yup, that should work fine

noggin/templates/user-settings-profile.html
```diff
@@ -12,7 +12,7 @@
     {% set buttontext = _('Change Avatar')%}
     {%- if OPENID is defined %}
     <form method="GET" action="https://www.libravatar.org/openid/login/">
-      <input type="hidden" name="openid_identifier" value="{{ 'https://'+user.username+OPENID }}">
+      <input type="hidden" name="openid_identifier" value="{{ OPENID % format(user.username) }}">
```
abompard: This crashes Jinja, it's either `OPENID % user.username` or `OPENID|format(user.username)`. The [docs](https://jinja.palletsprojects.com/en/2.11.x/templates/#format) suggest the former.

abompard (changes requested):

noggin/templates/user-settings-profile.html
Conan-Kudo: This should be fixed now.

abompard (approved):

PR #542, opened 2021-03-22
Added instruction to make a dir fixing vagrant not starting

copperii:
'vagrant up' does not start without this directory, gives on error:
* The host path of the shared folder is missing: ../freeipa-fas


copperii:
@abompard fixed it, no longer needed.

PR #545, opened 2021-03-23
Add unallowed usernames check

### LINKED ISSUE #512: allow for configurable blocklist of account names
In fas we had a list of usernames that were disallowed to be created. 
Would be great to have that in noggin as well. 

You can see the list in: 
https://pagure.io/fedora-infra/ansible/blob/main/f/roles/fas_server/templates/fas.cfg.j2#_76

Rationale:
* Sometimes we create email aliases, we don't want someone to make a user with the same name and never be able to get their email or confuse the alias people (examples: fpl@fedoraproject.org going to the fedora project leader, etc)
* We don't want users to create accounts that overlap with system accounts, as this could give them privs or cause issues with system accounts (examples: root, sync, ftp, bodhi, kojibuilder, etc)

 Likely in the new noggin world we would want to either fold the CentOS list into ours.

copperii:
Fixes #512

### DIFF
--- a/noggin/controller/registration.py
+++ b/noggin/controller/registration.py
@@ -91,6 +91,12 @@ def handle_register_form(form):
         .replace("  ", " ")
         .strip()
     )
+    # Here we should check if a username is allowed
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
@copperii Can you rebase this on current `dev` branch? There's a lot of unrelated commits in here...

copperii:
@Conan-Kudo it is now rebased

Conan-Kudo:
That's not exactly what I expected...

Can you try instead the following?

```
git remote add upstream https://github.com/fedora-infra/noggin.git
git fetch --all
git rebase upstream/dev
git push -f
```

copperii:
@Conan-Kudo did as you asked.

Conan-Kudo:
@copperii Can you drop the "rebase branch" commit?

```
git rebase -i HEAD~2
<in the editor, tell it to drop rebase branch commit>
git push -f
```

Conan-Kudo:
Perfect!

Conan-Kudo:
@copperii It looks like your commit only changes the tests? Where's the code for the rest of it?

noggin/controller/registration.py
```diff
@@ -91,6 +91,12 @@ def handle_register_form(form):
...
...p","dgilmore","directors","dovecot","dumper","fama","famsco","fas","fas_sync","fax","fcaic","fedora","fedorarewards","fedora-admin-xmlrpc","fesco","freemedia","freshmaker","ftbfs","ftp","ftpadm","ftpadmin","ftpsync","games","gdm","git","gnomebackup","gopher","gregdek","grokmirror","halt","hostmaster","hotness","ident","info","ingres","jaboutboul","jan","jwf","keys","kojiadmin","ldap","legal","logo","lp","m8y","mail","mailnull","manager","marketing","masher","masta","mirrormanager","mysql","nagios","named","netdump","news","newsadm","newsadmin","nfsnobody","nobody","noc","notallowed","notifications","nrpe","nscd","ntp","nut","openvideo","operator","packager","pagure","paguremirroring","patrick","pcap","pkgdb","pkgsigner","postfix","postgres","postmaster","press","privoxy","pvm","quagga","radiusd","radvd","relnotes","relrod","rel-eng","root","rpc","rpcuser","rpm","rsc","s3-mirror","sales","scholarship","secalert","secondary-signer","security","server-wg","shutdown","smmsp","spevack","squid","sshd","support","sync","system","temp","test","tickets","toor","updates","usenet","uucp","vcsa","vendors","vendor-support","voting","webalizer","webmaster","wikiadmin","wnn","www","xfs","zabbix"}
```
Conan-Kudo, line 95: Shouldn't this be in the noggin configuration?

Conan-Kudo (changes requested):

copperii:
@Conan-Kudo got dropped somewhere, restored.

noggin/controller/registration.py
copperii: Could be a better place, is defaults.py ok?

abompard:
Hey! So the solution that we decided on in the end is to just create the forbidden users in IPA directly (as disabled accounts). This way noone will be able to create a user with the same name. The good thing about this method is that IPA stays the reference for as much information as we can, instead of configuration being scattered among app configuration files.

As a result, we won't need this feature in Noggin, but thanks for proposing it! I hope you got to learn a bit about how the code is layed out and it'll make future contributions easier.

copperii:
Ok, thanks.

PR #549, opened 2021-03-25
Add an additional confirmation when closing OTP QR

ryanlerch:
Adds an additional confirmation when a user tries to close the OTP
QRcode, making sure that they understand that this is not going to be
shown again.

### DIFF
--- a/noggin/templates/user-settings-otp.html
+++ b/noggin/templates/user-settings-otp.html
@@ -114,6 +114,9 @@
         $('#otp-toggle').click(function() {
           $('#otp-qrcode').slideToggle("fast");
         })
+        $('#otp-modal').on("hide.bs.modal", function (e) {
+          if(!confirm("{{_('Are you sure you want to close the OTP dialog? This code will not be shown again, and you will be locked out of your account if you have no way to generate your OTP')}}")) return false;
+        });
     });
   </script>
   {% endif %}


Conan-Kudo (approved):

noggin/templates/user-settings-otp.html
```diff
@@ -114,6 +114,9 @@ <h5 id="pageheading">{{ _("OTP Tokens") }}</h5>
         $('#otp-toggle').click(function() {
           $('#otp-qrcode').slideToggle("fast");
         })
+        $('#otp-modal').on("hide.bs.modal", function (e) {
+          if(!confirm("{{_('Are you sure you want to close the OTP dialog? This code cannot be shown again, and you will be locked out of your account if you have no way to generate your an OTP')}}")) return false;
```
abompard: > and you will be locked out of your account if you have no way to generate your an OTP

I'm not parsing the end of that sentence right.

ryanlerch: thanks! wording updated. It's still meh, but at least parses. :)

noggin/templates/user-settings-otp.html
```diff
@@ -114,6 +114,9 @@ <h5 id="pageheading">{{ _("OTP Tokens") }}</h5>
         $('#otp-toggle').click(function() {
           $('#otp-qrcode').slideToggle("fast");
         })
+        $('#otp-modal').on("hide.bs.modal", function (e) {
+          if(!confirm("{{_('Are you sure you want to close the OTP dialog? This code will not be shown again, and you will be locked out of your account if you have no way to generate your OTP')}}")) return false;
```
abompard, line 118: Instead of doing `if (!something()) return false;` I suppose you could just do `return something();`. Or if you really worry about types, you could do `return !!something();`. It looks ugly but it's apparently the standard way to cast to bool in javascript.

nirik:
I wonder if perhaps we could make them/ask them to 'test' their otp here? ie, enroll, it puts up the qr code, they enroll it and then it asks them to enter the otp to confirm that it's valid and that they have it working, then the qr code and popup goes away. 
That way we are sure they actually have it working before we dismiss the qrcode?

abompard:
This is superseded by #584 (implementing Kevin's suggestion).

PR #569, opened 2021-04-02
fade out successfull flash messages in 10 seconds

copperii:
Successful messages are annoyingly in front when scrolling the page, this hides them automatically in 10 seconds if not closed before that.


ryanlerch:
Thanks for filing this, and sorry about the slow feedback loop.

This is a great idea, however:

We recently migrated to fedora bootstrap 2 (bootstrap 5) - and the "toasts" in there can be configured easily to dissapear over time.

I just filed #1380 to use toasts

PR #581, opened 2021-04-12
Making Group's mailing_list a link

### LINKED ISSUE #552: Links to a group's mailing list are broken
This afternoon I noticed that links to a group's mailing list were broken.

If you go to https://accounts.fedoraproject.org/group/sysadmin/ and click on the link next to the 📧 icon, that says infrastructure[at]lists.fedoraproject.org, you will be brought to https://accounts.fedoraproject.org/group/sysadmin/infrastructure@lists.fedoraproject.org, which 404s.

[In FAS2](https://admin.fedoraproject.org/accounts/group/view/sysadmin), the comparable link would go to the proper location, which is the list info page. This allows a user to view the mailing list history and subscribe to the list.

When addressing this bug, please note that there are groups that host their mailing lists on systems other than lists.fedoraproject.org, so the fix is not as simple as concatenating the hyperkitty URL to the group name. Take, for example, [fedora-hams](https://accounts.fedoraproject.org/group/fedora-hams/) who hosts their mailing list on the kernelnewbies.org mailing list system.

jossperdomo:
Fixes #552 

Added `mailto` in group template to `mailing_list` attribute.

To make a full replica of FAS system a new attribute `mailing_list_url` would be added to LDAP system and then added to `Group` model, form and template.

### DIFF
--- a/noggin/templates/group.html
+++ b/noggin/templates/group.html
@@ -18,7 +18,7 @@
           </h3>
           {% if group.description %}<div>{{ group.description }}</div>{% endif %}
           {% if group.mailing_list %}
-            <div id="group-mailinglist"><i class="fa fa-fw fa-envelope" aria-hidden="true"></i> <a href="{{group.mailing_list}}">{{group.mailing_list}}</a></div>
+            <div id="group-mailinglist"><i class="fa fa-fw fa-envelope" aria-hidden="true"></i> <a href="mailto:{{group.mailing_list}}">{{group.mailing_list}}</a></div>
           {% endif %}
           {% if group.irc_channel %}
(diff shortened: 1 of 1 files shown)


jossperdomo:
Hello @abompard and @phuzion! After setting the dev environment and check the code and your [FAS schema](https://github.com/fedora-infra/freeipa-fas/blob/dev/schema.d/89-fasschema.ldif) I figured out that added a `mailing_list_url` depends on this LDAP schema and the solution proposed by @phuzion would be a little bit more complex to implement.

Please let me know any comments 😊

abompard (approved):

PR #587, opened 2021-04-14
Separate UI field for otp token

### LINKED ISSUE #572: Use seperate UI field for otp token
Currently the noggin login page has just a username/password username/password+otp pair of fields. 
It would be nice to split the otp out and make it it's own (optional) field. 
This would help folks that use web password managers, and make the UI a little nicer. :)

jossperdomo:
Resolves #572 

Add a separate field in the login form for `one-time-password`. 
Including a new test case with login using password + OTP. 

image

### DIFF
--- a/noggin/form/edit_user.py
+++ b/noggin/form/edit_user.py
@@ -115,7 +115,7 @@ class UserSettingsAddOTPForm(ModestForm):
     otp = PasswordField(
         _('One-Time Password'),
         validators=[Optional()],
-        description=_("Enter your One-Time Password token if you have enrolled one"),
+        description=_("Enter your One-Time Password"),
     )
 
     submit = SubmitButtonField(_("Generate OTP Token"))
--- a/noggin/form/edit_user.py
+++ b/noggin/form/edit_user.py
@@ -115,7 +115,7 @@ class UserSettingsAddOTPForm(ModestForm):
     otp = PasswordField(
         _('One-Time Password'),
         validators=[Optional()],
-        description=_("Enter your One-Time Password token if you have enrolled one")
+        description=_("Enter your One-Time Password token if you have enrolled one"),
     )
 
     submit = SubmitButtonField(_("Generate OTP Token"))
--- a/noggin/form/edit_user.py
+++ b/noggin/form/edit_user.py
@@ -112,6 +112,12 @@ class UserSettingsAddOTPForm(ModestForm):
         description=_("please reauthenticate so we know it is you"),
     )
 
+    otp = PasswordField(
+        _('One-Time Password'),
+        validators=[Optional()],
+        description=_("Enter your One-Time Password token if you have enrolled one")
+    )
+
     submit = SubmitButtonField(_("Generate OTP Token"))
 
 
--- a/noggin/form/password_reset.py
+++ b/noggin/form/password_reset.py
@@ -19,11 +19,7 @@ class NewPasswordForm(BaseForm):
 
     password_confirm = PasswordField(_('Confirm New Password'))
 
-    otp = PasswordField(
-        _('One-Time Password'),
-        validators=[Optional()],
-        description=_("Enter your One-Time Password (if you have one)"),
-    )
+    otp = StringField(_('One-Time Password'), validators=[Optional()])
 
 
 class PasswordResetForm(NewPasswordForm):
@@ -31,7 +27,6 @@ class PasswordResetForm(NewPasswordForm):
     current_password = PasswordField(
         _('Current Password'),
         validators=[DataRequired(message=_('Current password must not be empty'))],
-        description=_("Just the password, don't add the OTP token if you have one"),
     )
 
 
--- a/noggin/form/password_reset.py
+++ b/noggin/form/password_reset.py
@@ -22,7 +22,7 @@ class NewPasswordForm(BaseForm):
     otp = PasswordField(
         _('One-Time Password'),
         validators=[Optional()],
-        description=_("Enter your One-Time Password token if you have enrolled one"),
+        description=_("Enter your One-Time Password (if you have one)"),
     )
 
 
--- a/noggin/form/password_reset.py
+++ b/noggin/form/password_reset.py
@@ -1,6 +1,6 @@
 from flask_babel import lazy_gettext as _
 from wtforms import PasswordField, StringField
-from wtforms.validators import DataRequired, EqualTo
+from wtforms.validators import DataRequired, EqualTo, Optional
 
 from .base import BaseForm
 from .validators import PasswordLength
@@ -19,8 +19,10 @@ class NewPasswordForm(BaseForm):
 
     password_confirm = PasswordField(_('Confirm New Password'))
 
-    otp = StringField(
-        _('OTP Token'), description=_("Enter your OTP token if you have enrolled one")
+    otp = PasswordField(
+        _('One-Time Password'),
+        validators=[Optional()],
+        description=_("Enter your One-Time Password token if you have enrolled one"),
     )
 
 
--- a/noggin/templates/_login_form.html
+++ b/noggin/templates/_login_form.html
@@ -5,8 +5,11 @@
     <div class="form-group">
       {{ macros.with_errors(login_form.username, class="validate", placeholder="Username", tabindex="1", label=False) }}
     </div>
+    <div class="form-group">
+      {{ macros.with_errors(login_form.password, class="validate", placeholder="Password", tabindex="2", label=False) }}
+    </div>
     <div class="form-group mb-0">
-      {{ macros.with_errors(login_form.password, class="validate", placeholder="Password or Password + One-Time-Password", tabindex="2", label=False) }}
+      {{ macros.with_errors(login_form.otp, class="validate", placeholder="One-Time-Password (optional)", tabindex="2", label=False) }}
     </div>
     <div class="form-group mb-0 text-right">
     {% if lost_otp_token is not defined %}
--- a/noggin/templates/_login_form.html
+++ b/noggin/templates/_login_form.html
@@ -9,7 +9,7 @@
       {{ macros.with_errors(login_form.password, class="validate", placeholder="Password", tabindex="2", label=False) }}
     </div>
     <div class="form-group mb-0">
-      {{ macros.with_errors(login_form.otp, class="validate", placeholder="One-Time-Password (optional)", tabindex="2", label=False) }}
+      {{ macros.with_errors(login_form.otp, class="validate", placeholder="One-Time Password (optional)", tabindex="2", label=False) }}
     </div>
     <div class="form-group mb-0 text-right">
     {% if lost_otp_token is not defined %}
--- a/noggin/templates/_login_form.html
+++ b/noggin/templates/_login_form.html
@@ -9,7 +9,7 @@
       {{ macros.with_errors(login_form.password, class="validate", placeholder="Password", tabindex="2", label=False) }}
     </div>
     <div class="form-group mb-0">
-      {{ macros.with_errors(login_form.otp, class="validate", placeholder="One-Time Password (optional)", tabindex="3", label=False) }}
+      {{ macros.with_errors(login_form.otp, class="validate", placeholder="One-Time Password (if you have one)", tabindex="3", label=False) }}
     </div>
     <div class="form-group mb-0 text-right">
     {% if lost_otp_token is not defined %}
--- a/noggin/templates/_login_form.html
+++ b/noggin/templates/_login_form.html
@@ -9,7 +9,7 @@
       {{ macros.with_errors(login_form.password, class="validate", placeholder="Password", tabindex="2", label=False) }}
     </div>
     <div class="form-group mb-0">
-      {{ macros.with_errors(login_form.otp, class="validate", placeholder="One-Time Password (optional)", tabindex="2", label=False) }}
+      {{ macros.with_errors(login_form.otp, class="validate", placeholder="One-Time Password (optional)", tabindex="3", label=False) }}
     </div>
     <div class="form-group mb-0 text-right">
     {% if lost_otp_token is not defined %}
@@ -23,6 +23,6 @@
   </div>
   <div class="card-footer d-flex justify-content-between">
     <div>{{ macros.non_field_errors(login_form) }}</div>
-    {{ login_form.submit(color="primary", tabindex="3") }}
+    {{ login_form.submit(color="primary", tabindex="4") }}
   </div>
 </form>
--- a/noggin/templates/user-settings-otp.html
+++ b/noggin/templates/user-settings-otp.html
@@ -63,7 +63,7 @@
           <div class="form-group">{{ macros.with_errors(addotpform.description, label=False, tabindex="2", placeholder=_("Token name"))}}</div>
           <div class="form-group">{{ macros.with_errors(addotpform.password, label=False, tabindex="3", placeholder=_("Password"))}}</div>
           {% if tokens %}
-          <div class="form-group">{{ macros.with_errors(addotpform.otp, label=False, tabindex="4", placeholder=_("One-Time Password (optional)"))}}</div>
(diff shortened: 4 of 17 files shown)


ryanlerch:
I think there are a few other places where we input a password. (creating a new otp token is somewhere that comes to mind), and there may be a handful of others too

noggin/form/password_reset.py
```diff
@@ -19,8 +19,10 @@ class NewPasswordForm(BaseForm):
 
     password_confirm = PasswordField(_('Confirm New Password'))
 
-    otp = StringField(
-        _('OTP Token'), description=_("Enter your OTP token if you have enrolled one")
+    otp = PasswordField(
```
abompard: I think this can remain a StringField, the value is temporary so no need to hide it.

noggin/templates/_login_form.html
```diff
@@ -5,8 +5,11 @@
     <div class="form-group">
       {{ macros.with_errors(login_form.username, class="validate", placeholder="Username", tabindex="1", label=False) }}
     </div>
+    <div class="form-group">
+      {{ macros.with_errors(login_form.password, class="validate", placeholder="Password", tabindex="2", label=False) }}
+    </div>
     <div class="form-group mb-0">
-      {{ macros.with_errors(login_form.password, class="validate", placeholder="Password or Password + One-Time-Password", tabindex="2", label=False) }}
+      {{ macros.with_errors(login_form.otp, class="validate", placeholder="One-Time Password (optional)", tabindex="2", label=False) }}
```
abompard: Please increment the `tabindex`.

noggin/templates/user-settings-otp.html
```diff
@@ -46,10 +46,9 @@ <h5 class="modal-title">{{ _("Add OTP Token") }}</h5>
         <form action="{{ url_for('.user_settings_otp', username=current_user.username) }}" method="post" class="px-4 py-3" novalidate>
           <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
           <div class="form-group">{{ macros.with_errors(addotpform.description, label=False, tabindex="2", placeholder=_("Token name"))}}</div>
-          {% if tokens %}
-          <div class="form-group">{{ macros.with_errors(addotpform.password, label=False, tabindex="3", placeholder=_("Password + One-Time-Password"))}}</div>
-          {% else %}
           <div class="form-group">{{ macros.with_errors(addotpform.password, label=False, tabindex="3", placeholder=_("Password"))}}</div>
+          {% if tokens %}
+          <div class="form-group">{{ macros.with_errors(addotpform.otp, label=False, tabindex="3", placeholder=_("One-Time Password (optional)"))}}</div>
```
abompard: Please increment the `tabindex` here too.

noggin/tests/unit/controller/test_authentication.py
```diff
@@ -91,6 +91,27 @@ def test_login(client, dummy_user):
...
 
+@pytest.mark.vcr()
+def test_login_with_otp(client, dummy_user):
+    """Test a successful Login with password + otp"""
+    result = client.post(
+        '/',
+        data={
+            "login-username": "dummy",
+            "login-password": "dummy_password",
+            "login-otp": "123456",
```
abompard: Hmm, this should not work, because the dummy user does not have an OTP token enabled by default in tests. You have to use the `dummy_user_with_otp` fixture instead of the `dummy_user` fixture. You can look at the tests in the  `test_user_otp.py` module to see how you can get the current value of the token for your otp form field.

I suppose you copied the cassette and edited the content but that's not how cassettes are generated, it's much simpler: just run the unit tests inside the vagrant VM and the cassette will be automatically created.

abompard (changes requested):
Thanks for your contribution! There are a few necessary changes but it's good quality overall :+1:

jossperdomo:
> I think there are a few other places where we input a password. (creating a new otp token is somewhere that comes to mind), and there may be a handful of others too

Yes, you're right. After opening this PR, I'd figured out this field is also used to adding a new OTP token, I'm adding it :)

noggin/tests/unit/controller/test_authentication.py
jossperdomo: Yes, I didn't know how VCR works. I've searched about it and after discovering the `--vcr-record` flag I finally was able to create news cassettes 😊 

I have some doubts related to how to make a not logged user with an OTP token to test it because after so many tries I failed to create a new `dummy_user`. The current `dummy_user_with_otp` is already logged and doesn't fit to test the login. I created a new fixture of a user (I refactored just a little bit the `make_user` function) and I added a new otp token after setting the password with `changed_password`: 

```
@pytest.fixture
def make_user(ipa_testing_config, app):
    created_users = []

    def _make_user(name, with_otp=False):
        now = datetime.datetime.utcnow().replace(microsecond=0)
        password = f'{name}_password'
        ipa_admin.user_add(
            a_uid=name,
            o_givenname=name.title(),
            o_sn='User',
            o_cn=f'{name.title()} User',
            o_mail=f"{name}@example.com",
            o_userpassword=password,
            o_loginshell='/bin/bash',
            fascreationtime=f"{now.isoformat()}Z",
        )
        created_users.append(name)
        ipa = untouched_ipa_client(app)
        ipa.change_password(name, new_password=password, old_password=password)

        if with_otp:
            otp_token = ipa_admin.otptoken_add(o_ipatokenowner=name, o_description=f"{name}'s token")
            otp_token = OTPToken(otp_token["result"])
            return otp_token

    yield _make_user

    for name in created_users:
        ipa_admin.user_del(name)
```

When I test the login sending the OTP code (just like the following example) I get `invalid-password` error from the login and (of course) the test fails.

```
result = client.post(
        '/',
        data={
            "login-username": "dummy",
            "login-password": "dummy_password",
            "login-otp": otp,
            "login-submit": "1",
        },
        follow_redirects=True,
    )
```
Could you please tell me if the connection to the IPA with the token should be mocked?

abompard: > The current `dummy_user_with_otp` is already logged and doesn't fit to test the login

You're right! So what I think you should do is either:
- create a fixture that is similar to what `dummy_user_with_otp` does except it takes a `dummy_user` as dependency (the function argument) instead of a `logged_in_dummy_user`, and adds the token with `ipa_admin` instead of `ipa`
- or add the token inside your test function, but use the `cleanup_dummy_tokens` fixture to remove it afterwards

I recommend doing composition with fixtures (new fixtures which depend on other fixtures) rather than refactoring existing fixtures to add more features.

Then, in your test, you can do something like `current_otp = get_otp(otp_secret_from_uri(token.uri))` to get the current value of the token and add it to your POST.

Feel free to ask me questions if you're blocked.

jossperdomo: Hello again! I'm still struggling to finish successfully the test. I did the fixture just like you said but when I run the test, in fact when I run the tests recording the cassette in the VM, the ones which use a user with tokens added will fail. All the time these tests return [`kinit: Preauthentication failed while getting initial credentials`](https://lists.fedorahosted.org/archives/list/freeipa-users@lists.fedorahosted.org/message/XT5GZFGHVEQH2LH56UYC56EIXX2N6PTH/). 

This image shows an example of a failure. Instead of return the user profile page, it is the login page with the message of `bad credentials` added to password field in the login form. 
image

I updated the PR with my current code of the tests and not the cassettes. I'm quite sure the error comes from my freeipa local VM, could you please can help me to configure the freeipa VM to avoid this error? I'm sorry to bother you with this issue but I couldn't solve it by myself.

abompard: My mistake! We made a change in the hash algorithm used in OTP tokens a while ago and forgot to update the test functions. Please rebase on the HEAD of the dev branch, it should make this error disappear. Sorry about that!
And thanks for your perseverance.

noggin/templates/_login_form.html
```diff
@@ -5,8 +5,11 @@
     <div class="form-group">
       {{ macros.with_errors(login_form.username, class="validate", placeholder="Username", tabindex="1", label=False) }}
     </div>
+    <div class="form-group">
+      {{ macros.with_errors(login_form.password, class="validate", placeholder="Password", tabindex="2", label=False) }}
+    </div>
     <div class="form-group mb-0">
-      {{ macros.with_errors(login_form.password, class="validate", placeholder="Password or Password + One-Time-Password", tabindex="2", label=False) }}
+      {{ macros.with_errors(login_form.otp, class="validate", placeholder="One-Time Password (optional)", tabindex="3", label=False) }}
```
abompard: Maybe "if you have one" would be better than "optional"?

noggin/templates/user-settings-otp.html
```diff
@@ -61,13 +61,12 @@ <h5 class="modal-title">{{ _("Add OTP Token") }}</h5>
         <form action="{{ url_for('.user_settings_otp', username=current_user.username) }}" method="post" class="px-4 py-3" novalidate>
           <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
           <div class="form-group">{{ macros.with_errors(addotpform.description, label=False, tabindex="2", placeholder=_("Token name"))}}</div>
-          {% if tokens %}
-          <div class="form-group">{{ macros.with_errors(addotpform.password, label=False, tabindex="3", placeholder=_("Password + One-Time-Password"))}}</div>
-          {% else %}
           <div class="form-group">{{ macros.with_errors(addotpform.password, label=False, tabindex="3", placeholder=_("Password"))}}</div>
+          {% if tokens %}
+          <div class="form-group">{{ macros.with_errors(addotpform.otp, label=False, tabindex="4", placeholder=_("One-Time Password (optional)"))}}</div>
```
abompard: The OTP token is not actually optional here: the field is only displayed if there's already an existing token, therefore the OTP field is mandatory.

noggin/tests/unit/controller/test_authentication.py
```diff
@@ -91,6 +93,28 @@ def test_login(client, dummy_user):
...
+        data={
+            "login-username": "dummy",
+            "login-password": "dummy_password",
+            "login-otp": otp,
+            "login-submit": "1",
+        },
+        follow_redirects=True,
+    )
+    page = BeautifulSoup(result.data, 'html.parser')
+    messages = page.select(".flash-messages .alert-success")
```
abompard: In the `tests.utilities` module, you'll find an `assert_redirects_with_flash()` function that would be appropriate here instead of parsing the HTML.

noggin/tests/unit/controller/test_password_reset.py
```diff
@@ -84,7 +84,9 @@ def test_password_changes_user(client, logged_in_dummy_user):
 
 
 @pytest.mark.vcr()
-def test_password_form_with_otp(client, logged_in_dummy_user, dummy_user_with_otp):
+def test_password_form_with_otp(
+    client, logged_in_dummy_user, logged_in_dummy_user_with_otp
```
abompard: I don't think you need `logged_in_dummy_user` here, since it's already contained in `logged_in_dummy_user_with_otp` (and the test does not use its value).

noggin/tests/unit/controller/test_user_otp.py
```diff
@@ -12,17 +12,19 @@
...
     assert_form_generic_error,
     assert_redirects_with_flash,
+    get_otp,
+    otp_secret_from_uri,
 )
 
 
 @pytest.fixture
-def dummy_user_with_2_otp(client, logged_in_dummy_user, dummy_user_with_otp):
+def dummy_user_with_2_otp(client, logged_in_dummy_user, logged_in_dummy_user_with_otp):
```
abompard, line 21: Same remark here.

noggin/tests/unit/controller/test_user_otp.py
```diff
@@ -147,7 +149,49 @@ def test_user_settings_otp_confirm(
...
+        confirm_form.select_one("input[name='confirm-secret']")["value"]
+        == parsed_otp_uri_query["secret"][0]
+    )
+
+
+@pytest.mark.vcr()
+def test_user_settings_otp_add_second_confirm(
+    client,
+    logged_in_dummy_user_with_otp,
+    logged_in_dummy_user,
```
abompard: Same remark here

noggin/tests/unit/controller/test_user_otp.py
```diff
@@ -387,12 +431,12 @@ def test_user_settings_otp_disable(client, logged_in_dummy_user, dummy_user_with
 
 @pytest.mark.vcr()
 def test_user_settings_otp_disable_lasttoken(
-    client, logged_in_dummy_user, dummy_user_with_otp
+    client, logged_in_dummy_user, logged_in_dummy_user_with_otp
```
abompard: Same remark here

noggin/tests/unit/translations/test_translations.py
```diff
@@ -18,13 +18,13 @@ def compile_catalogs():
 
 @pytest.mark.vcr()
 def test_translation_in_code_french(
-    client, logged_in_dummy_user, dummy_user_with_otp, compile_catalogs
+    client, logged_in_dummy_user, logged_in_dummy_user_with_otp, compile_catalogs
```
abompard: Same remark here

noggin/form/edit_user.py
```diff
@@ -111,6 +111,12 @@ class UserSettingsAddOTPForm(ModestForm):
         description=_("please reauthenticate so we know it is you"),
     )
 
+    otp = StringField(
+        _('One-Time Password'),
+        validators=[Optional()],
+        description=_("Enter your One-Time Password token if you have enrolled one"),
```
abompard: Actually, the field will only be displayed if there is an existing token, so you don't need "if you have enrolled one".

abompard (changes requested):
It's looking good! Just a few cleanups here and there.

noggin/tests/unit/controller/test_user_otp.py
jossperdomo: I think here it would be needed because `logged_in_dummy_user_with_otp` only returns the otp token parsed and a few lines below `logged_in_dummy_user` is used to add a new token. Maybe this fixture would be added in `conftest.py` or `logged_in_dummy_user_with_otp` would be improved later to returns the client handler with the token.

noggin/templates/_login_form.html
jossperdomo: Done! 😊  yes, it would be better.

noggin/tests/unit/controller/test_authentication.py
jossperdomo: Yes, I took `test_login` as a model and changed it just a little bit. I updated both with this function 😊

noggin/tests/unit/controller/test_user_otp.py
abompard: Correct, I missed that.

I agree that ideally `logged_in_dummy_user_with_otp` would return both the handler and the token. It'll be quite a few changes throughout the tests code, so it's great if you're willing to do that, but if you don't I won't hold this PR for that.

noggin/form/password_reset.py
```diff
@@ -19,17 +19,14 @@ class NewPasswordForm(BaseForm):
...
-    )
+    otp = StringField(_('One-Time Password'), validators=[Optional()])
 
 
 class PasswordResetForm(NewPasswordForm):
 
     current_password = PasswordField(
         _('Current Password'),
         validators=[DataRequired(message=_('Current password must not be empty'))],
-        description=_("Just the password, don't add the OTP token if you have one"),
```
abompard, line 32: This broke the `test_password_form_with_otp` test, so please adjust it.

abompard (changes requested):
Okay, getting there... :-) One test fails, and one optional refactoring work if you want to do it.

Please also add a news file as described in [the contributor docs](https://noggin-aaa.readthedocs.io/en/latest/contributing.html#release-notes) with the snippet you'd like to see in the release notes.

Also, you'll have to rebase on the current `dev` branch to fix conflicts. I can help with that if needed.

abompard (approved):
Looks great, thanks!

Issue #594, opened 2021-04-15
do not allow upper case login's to work


nirik:
noggin apparently lets me login as:
kevin
Kevin
KeVIN
However, ipsilon/ipa will fail to authenticate any of these except the all lower case one. 
We should probibly reject any logins with upper case and tell the user to try lower case...

nirik:
Seems ipa web also allows any mixed case... and 'ipa user-show KeVIN' matches 'kevin'
So, I guess it's just ipsilon here thats not working with mixed case...

abompard:
Yeah I'd rather lowercase the username field in Noggin than try to support mixed case in Ipsilon.

nirik:
Well, it's kind of a mess here, because if noggin just lowercases the login, but ipsilon doesn't, people are going to fail to authenticate. ;( 
I'm fine with lower casing everything, but it should be everything then. Perhaps there's some options with the ipa connector ipsilon is using?

jossperdomo:
Is this issue related to #573?

abompard:
No it's unrelated, I believe.

nirik:
Also, this breaks on password resets. 
If foobar tries to reset their password and they login as 'Foobar' noggin correctly sends them a reset token, but then when they click on it they get a 500 from noggin:

```
Traceback (most recent call last):
  File "/opt/app-root/lib/python3.6/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/opt/app-root/lib/python3.6/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/opt/app-root/lib/python3.6/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/opt/app-root/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/opt/app-root/lib/python3.6/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/opt/app-root/lib/python3.6/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/opt/app-root/src/noggin/controller/password.py", line 206, in forgot_password_change
    lock.delete()
  File "/opt/app-root/src/noggin/utility/password_reset.py", line 30, in delete
    os.remove(self._get_file_path())
[...]

PR #617, opened 2021-05-02
Lowercase the username in Forgot Password Ask controller 

### LINKED ISSUE #573: Lowercase the usernames when users want to reset their password
If a user uses a mixed-case username to request a password change token, the lock file will be created with this mixed-case username. The IPA API call will succeed because IPA automatically lowercases usernames. But when the lock needs to be removed in the next part of the process, it will try to remove the lowercase version, not find it, and traceback.

We should lowercase the user input before creating the `PasswordResetLock` instance.

jossperdomo:
Resolves #573 

- Convert to lowercase username before creating `PasswordResetLock`
- Added a unit test with a mixed-case username, it is different from the username used in the fixture.

### DIFF
--- a/noggin/controller/password.py
+++ b/noggin/controller/password.py
@@ -126,7 +126,7 @@ def user_settings_password(ipa, username):
 def forgot_password_ask():
     form = ForgottenPasswordForm()
     if form.validate_on_submit():
-        username = form.username.data
+        username = form.username.data.lower()
         lock = PasswordResetLock(username)
         valid_until = lock.valid_until()
         now = datetime.datetime.now()
--- a/noggin/controller/password.py
+++ b/noggin/controller/password.py
@@ -126,7 +126,7 @@ def user_settings_password(ipa, username):
 def forgot_password_ask():
     form = ForgottenPasswordForm()
     if form.validate_on_submit():
-        username = form.username.data.lower()
+        username = form.username.data
         lock = PasswordResetLock(username)
         valid_until = lock.valid_until()
         now = datetime.datetime.now()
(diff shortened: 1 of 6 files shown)


noggin/controller/password.py
```diff
@@ -126,7 +126,7 @@ def user_settings_password(ipa, username):
 def forgot_password_ask():
     form = ForgottenPasswordForm()
     if form.validate_on_submit():
-        username = form.username.data
+        username = form.username.data.lower()
```
abompard: This will work, but I'd prefer adding a filter in the Form definition itself. This way if the form is ever used elsewhere, we won't have this bug again.
You can check how some form fields already use the `filters` argument for an example, and WTForms has documentation about it too.

abompard (changes requested):

noggin/controller/password.py
jossperdomo: I added the new filter in `base.py` with the previous one, `strip`. Is it okay?

noggin/tests/unit/controller/test_forgot_password.py
```diff
@@ -93,6 +93,20 @@ def test_ask_post_non_existant_user(client):
...
+    result = client.post('/forgot-password/ask', data={"username": "DuMmY"})
+    assert_redirects_with_flash(
+        result,
+        expected_url="/",
+        expected_message=(
+            "An email has been sent to your address with instructions on how to reset "
+            "your password"
+        ),
+        expected_category="success",
+    )
```
abompard, line 108: Nice ! Could you also check that the lock file has been created with the lowercased username please? (look into `patched_lock`)

abompard (changes requested):
Thanks! One more thing: please add a news file too.

noggin/controller/password.py
abompard: Yes, that's perfect, thanks :-)

abompard (approved):
Looks good, thanks!

If you want to try a somewhat similar issue, check out #594 ;-)

Issue #647, opened 2021-05-25
Language clean up


StephenCoady:
There are some usages of the word master that we could remove, [here](https://github.com/fedora-infra/noggin/blob/dev/devel/ansible/roles/common/vars/main.yml#L4) and [here](https://github.com/fedora-infra/noggin/blob/dev/devel/ansible/roles/freeipa/tasks/main.yml#L27)

We also have [master.html](https://github.com/fedora-infra/noggin/blob/dev/noggin/templates/master.html) which we could rename, along with the references to it, [here](https://github.com/fedora-infra/noggin/blob/dev/noggin/themes/fas/templates/main.html#L1), [here](https://github.com/fedora-infra/noggin/blob/dev/noggin/themes/centos/templates/main.html#L1), [here](https://github.com/fedora-infra/noggin/blob/dev/noggin/themes/default/templates/main.html#L1) and [here](https://github.com/fedora-infra/noggin/blob/dev/noggin/themes/openSUSE/templates/main.html#L1)

We can also swap usage of the deprecated [whitelist_externals](https://github.com/fedora-infra/noggin/blob/dev/tox.ini) to [allowlist_externals](https://tox.readthedocs.io/en/latest/config.html#conf-allowlist_externals)

ADHIKSHA:
Will love to work on this issue!

abompard:
Fixed by #648

PR #648, opened 2021-05-25
Issue #647 changes

cwg83:
This is for issue #647  

The only thing I'm unsure about is that I renamed master.html to main_template.html


abompard:
Since there's already a template called `main.html`, the new name for the master template should probably be something else, to avoid confusion (although it's a bit inevitable I suppose). Maybe `base.html`?

cwg83:
Good call @abompard - I've made the suggested change

abompard (approved):
Looks good, thanks!

PR #664, opened 2021-06-14
Display agreements on users profiles

### LINKED ISSUE #576: As a registered user, I want to tell if another user has signed the Fedora Project Contributor Agreement (FPCA).
# Summary

It should be easy to tell if another Fedora Account has signed the FPCA from their account profile.

# Background

Often I need to see whether a new contributor has signed the FPCA, since this blocks them from contributing to any other part of the Fedora Linux ecosystem, and maybe the CentOS one now too. In FAS2, ~20% of the time I used it to check if a newcomer has signed the FPCA already for this reason. However, it is not obvious or clear for me to tell if another contributor has signed the FPCA. It is easier for me to query via FASJSON, which is not always easier than logging in from a browser.

# Details

This is prominent info, so it should be near the top. I'd love to see a design, like a green checkmark, or a "verified" style icon, to appear in the profile if they had signed.

Plus, we are asking people to sign a legal agreement, even if it is an innovative one. So, a cool badge to appear in your profile to indicate you have done this important step might be one more incentive (and cause people to ask how to get that to appear in their profile).

# Outcome

Easier to tell if a new contributor has already signed the FPCA or not

abompard:
Fixes: #576

### DIFF
--- a/news/576.feature
+++ b/news/576.feature
@@ -1,0 +1,1 @@
+Display on users' profiles the agreements they have signed
--- a/noggin/templates/user.html
+++ b/noggin/templates/user.html
@@ -55,7 +55,7 @@
                   </strong>
                   <div class="text-right">
                     {% for ircnick in user.ircnick %}
-                    <div class="mb-0 p-1 text-monospace">{{ ircnick }}</div>
+                    <div class="mb-0 py-1 text-monospace">{{ ircnick }}</div>
                     {% endfor %}
                   </div>
                 </li>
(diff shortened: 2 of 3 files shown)


abompard:
I've displayed it as the top item in the list of information items on the profile's left column:

Users agreements

@ryanlerch, any opinion on the UI? Should it be lower in the list? Somewhere else?

Conan-Kudo:
Do the agreements _always_ have short names? Because if it's possible to not have short names, it probably should go in the longer list section similar to the group list.

abompard:
> Do the agreements _always_ have short names? Because if it's possible to not have short names, it probably should go in the longer list section similar to the group list.

Correct, it looks ugly with a longer name. Which is what we have in prod. @ryanlerch  any suggestions?

ryanlerch:
how about something like this?

image

Conan-Kudo:
That looks a lot better!

abompard:
Looks much better, thanks Ryan. I didn't find the exact same icon, so I picked another one. Where did you find this one?
Also, agreements don't have a description right now so it does not look exactly the same, but here it is:

profile

PR #668, opened 2021-06-18
Make the IRC nickname attribute not only about IRC

### LINKED ISSUE #248: As a registered community member, I want to be able to indicate my Matrix IDs in my account
In the same vein that we have for IRC nicks, there's a need to be able to indicate Matrix IDs (MXIDs).

Per [the Matrix specification](https://matrix.org/docs/spec/#users), the handle format is `@user:server`. The [full specification of the identifier format](https://matrix.org/docs/spec/appendices#identifier-grammar) should probably be implemented to validate whether an entered MXID is valid or not and to support the proper uniqueness constraints.

This is a requirement for openSUSE, as we're going to be standing up a Matrix server and "rehome" all community channels on it. Consequently, primacy will be moving from IRC to the Matrix, and we need a way to map member presence in the Matrix to accounts in the same way Fedora does with IRC.

This will also be useful for Fedora, as Matrix-only rooms for Fedora are starting to exist, and having the mapping would be very useful there, too.

abompard:
This will allow users to declare their Matrix IDs too.
   
Fixes: #248

### DIFF
--- a/news/248.feature
+++ b/news/248.feature
@@ -1,0 +1,1 @@
+Allow users to declare their Matrix IDs in addition to the IRC nicknames
--- a/noggin/app.py
+++ b/noggin/app.py
@@ -14,6 +14,7 @@ from noggin.middleware import IPAErrorHandler
 from noggin.security.ipa_admin import IPAAdmin
 from noggin.themes import Theme
 from noggin.utility import import_all
+from noggin.utility.templates import format_nickname
 
 
 # Forms
@@ -86,6 +87,10 @@ def create_app(config=None):
         },
         content_security_policy_nonce_in=['script-src'],
     )
+
+    # Template filters
+    # If there are too many, group them in an extension
+    app.jinja_env.filters["nickname"] = format_nickname
 
     # Register views
     import_all("noggin.controller")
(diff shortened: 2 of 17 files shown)


nirik:
This doesn't do anything toward validating them right? just allows for entering them in there?

abompard:
There is some validation, the field must have some format depending on the type of network chosen. For IRC it can be `username` or `username:server.name` or `username@server.name` and for Matrix it can be `username` or `username:server.name` or `@username:server.name`.

I could add more formats if you think of others.

Conan-Kudo:
The only valid format that should be permitted for Matrix is `@username:server` (the `@` is significant for MXIDs).

abompard:
@Conan-Kudo Since it's the nickname field of a user, we know it describes a username and not a room name, right? Also, I think it would be nice to let users omit the server name part, don't you think?

@nirik Did you ask the question because you think validating the fields would be a bad thing? Or have downsides?

Conan-Kudo:
> @Conan-Kudo Since it's the nickname field of a user, we know it describes a username and not a room name, right? Also, I think it would be nice to let users omit the server name part, don't you think?

If we visually indicate somehow what the default is when they fill it out, that's fine.

nirik:
> @nirik Did you ask the question because you think validating the fields would be a bad thing? Or have downsides?

No, validating is fine, I'd just love to verify that that nick/matrix id belongs to that user. But this is a good step forward until we can get verification.

abompard:
@Conan-Kudo : Here's how I made it look:

screenshot

Does this seem clear enough?

Conan-Kudo:
Looks good to me.

PR #672, opened 2021-06-25
Add a page to manage registering users

abompard:
This page supercedes the `spammish` application that we started before. It makes more sense to integrate it into Noggin, and is not less secure as I previously thought.

Activating this feature requires setting up a privilege and a role in IPA, as described in the documentation.


abompard:
Here's what it looks like:
screenshot

Conan-Kudo (approved):

PR #766, opened 2021-10-01
Improve the display of group communication channels

### LINKED ISSUE #309: As a community group admin, I want to indicate my group's Matrix room ID for Matrix users to join
This is really the community group variant of #248.  In the same vein that we have for IRC channels, there's a need to be able to indicate Matrix room IDs.

Per [the Matrix specification](https://matrix.org/docs/spec/#users), the room ID format is `#room:server`. The [full specification of the identifier format](https://matrix.org/docs/spec/appendices#identifier-grammar) should probably be implemented to validate whether an entered Matrix room ID is valid or not and to support the proper uniqueness constraints.

This is a requirement for openSUSE, as we've stood up a Matrix server and "rehomed" all community channels on it. Consequently, primacy has moved from IRC to the Matrix, and we need a way to map group room presence in the Matrix in the same way Fedora does with IRC.

This will also be useful for Fedora, as Matrix-only rooms for Fedora are starting to exist, and having the ability to set and surface this to users will allow them the ability to join from a Matrix client easily.

abompard:
IRC and Matrix are supported.

Fixes: #309

Noggin does not allow group sponsors to edit the group channel (yet), it has to go through an infra ticket, so we can't validate the input yet. Several formats are supported, as with nicknames.

### DIFF
--- a/news/309.feature
+++ b/news/309.feature
@@ -1,0 +1,1 @@
+Improve the display of group communication channels (IRC or Matrix)
--- a/noggin/app.py
+++ b/noggin/app.py
@@ -15,7 +15,7 @@ from noggin.middleware import IPAErrorHandler
 from noggin.security.ipa_admin import IPAAdmin
 from noggin.themes import Theme
 from noggin.utility import import_all
-from noggin.utility.templates import format_nickname
+from noggin.utility.templates import format_channel, format_nickname
 
 
 # Forms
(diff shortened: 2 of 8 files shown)


abompard:
This is what it looks like:

* With an IRC channel
Capture d’écran de 2021-10-04 12-22-52

* With a matrix room
Capture d’écran de 2021-10-04 12-36-57

PR #918, opened 2022-05-11
Fix some dependencies

gridhead:
Signed-off-by: Akashdeep Dhar <akashdeep.dhar@gmail.com>

Except for two, https://github.com/fedora-infra/noggin/issues/899#issuecomment-1123188847 and https://github.com/fedora-infra/noggin/issues/899#issuecomment-1123196803.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,16 +46,16 @@ classifiers = [
 python = "^3.9.0"
 cryptography = ">=3.4.7"
 flask = "^2.1.0"
-flask-wtf = "^1.0.0"
+flask-wtf = ">=0.14.3, <2.0.0"
 python-freeipa = "^1.0.6"
 click = "^8.0"
-pyjwt = "^2.3.0"
+pyjwt = "^2.1.0"
 Flask-Mail = "^0.9.1"
 gunicorn = {version = "^20.0", optional = true}
-fedora-messaging = "^3.0.0"
-backoff = "^2.0.0"
+fedora-messaging = ">=2.1.0, <4.0"
+backoff = ">=1.10.0, <3.0"
 noggin-messages = "^1.0.0"
-whitenoise = "^6.0.0"
+whitenoise = ">=5.3, <7.0"
 flask-babel = "^2.0.0"
 flask-healthz = "^0.0.3"
 markupsafe = "^2.0.0"
@@ -63,8 +63,8 @@ wtforms = {extras = ["email"], version = "^3.0.0"}
 requests = {extras = ["security"], version = "^2.26.0"}
 translitcodec = "0.7.0"
 unidecode = "^1.2.0"
-flask-talisman = "^1.0.0"
-pyotp = "^2.6.0"
+flask-talisman = ">=0.8.1, <2.0"
+pyotp = "^2.2.7"
 
 [tool.poetry.dev-dependencies]
 pytest = "^7.1.0"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -45,26 +45,26 @@ classifiers = [
 [tool.poetry.dependencies]
 python = "^3.9.0"
 cryptography = "^37.0.0"
-flask = "^2.1.0"
-flask-wtf = "^1.0.0"
+flask = "^2.0.1"
+flask-wtf = "^0.14.3"
```
abompard: We do want to support 1.0.0, so it should be something like `">=0.14.3, <2.0.0"`

pyproject.toml
```diff
@@ -45,26 +45,26 @@ classifiers = [
...
 click = "^8.0"
-pyjwt = "^2.3.0"
+pyjwt = "^2.1.0"
 Flask-Mail = "^0.9.1"
 gunicorn = {version = "^20.0", optional = true}
-fedora-messaging = "^3.0.0"
-backoff = "^2.0.0"
-noggin-messages = "^1.0.0"
-whitenoise = "^6.0.0"
+fedora-messaging = "^2.1.0"
```
abompard: We do want to support 3.0.0, so it should be something like `">=2.1.0, <4.0"`

pyproject.toml
```diff
@@ -45,26 +45,26 @@ classifiers = [
...
+backoff = "^1.10.0"
```
abompard: Same here, `">=1.10.0, <3.0"`

pyproject.toml
```diff
@@ -45,26 +45,26 @@ classifiers = [
...
+noggin-messages = "^0.0.1"
```
abompard: Nah, it should really be `"^1.0.0"` here, let's just package it since it's alongside this repo.

pyproject.toml
```diff
@@ -45,26 +45,26 @@ classifiers = [
...
+whitenoise = "^5.3.0"
```
abompard: Same here, we want to support 6.0.0, so it should be something like `">=5.3, <7.0"`

pyproject.toml
```diff
@@ -45,26 +45,26 @@ classifiers = [
...
 flask-babel = "^2.0.0"
 flask-healthz = "^0.0.3"
 markupsafe = "^2.0.0"
 wtforms = {extras = ["email"], version = "^3.0.0"}
 requests = {extras = ["security"], version = "^2.26.0"}
 translitcodec = "0.7.0"
 unidecode = "^1.2.0"
-flask-talisman = "^1.0.0"
-pyotp = "^2.6.0"
+flask-talisman = "^0.8.1"
```
abompard: Same here, `">=0.8.1, <2.0"`

abompard (changes requested):

pyproject.toml
gridhead: I will package `noggin-messages` first then, as this repo requires that to as a dependency.

gridhead:
I will let the jobs run their course while I start digging into `noggin-messages`. Also, the PR is now rebased on the most recent `dev`.

gridhead:
One of the checks here happens to fail here due to similar reasons that https://github.com/fedora-infra/fedora-messaging/pull/263 fails for.

abompard:
It should be fixed in dev now, please rebase.

abompard (approved):

PR #949, opened 2022-06-09
Remove Fedora-specific defaults from Noggin

Conan-Kudo:
This makes it easier for third-party (non-Fedora) deployments of Noggin to work.

### DIFF
--- a/news/PR949.bic
+++ b/news/PR949.bic
@@ -1,0 +1,1 @@
+Noggin no longer assumes it is being deployed in Fedora infrastructure by default.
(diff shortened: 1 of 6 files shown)


abompard (approved):

abompard:
Well, I would have set the news type as `bic` (backwards incompatible), because it's a default configuration change.

abompard (changes requested):
The news type should be bic (backwards incompatible), because it's a default configuration change.

Also please allow changes to your branch so that mergify can rebase. Thanks!

Conan-Kudo:
@abompard Done.

abompard (approved):

PR #958, opened 2022-06-21
Add a blocklist for registering users

### LINKED ISSUE #957: Add blocklist of account name substrings
Downstream report: 
https://pagure.io/fedora-infrastructure/issue/10743

This is asking for a way to prevent users from making accounts with 'fedora' in them, but I think this might be more generally useful as well if we could implement a list or file of them. 

So, basically a file or config key that lists substrings that are disallowed in account names.

abompard:
Fixes: #957

### DIFF
--- a/news/957.feature
+++ b/news/957.feature
@@ -1,0 +1,1 @@
+Add a blocklist for registering users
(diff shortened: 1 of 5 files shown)


abompard:
Hey @LenkaSeg , any idea why the "testing farm" checks fail?

LenkaSeg:
> Hey @LenkaSeg , any idea why the "testing farm" checks fail?

Looks like an issue on the testing farm: https://status.testing-farm.io/affected/public-ranch/

LenkaSeg:
Update: I got a suggestion from @TomasTomecek to remove the testing-farm for the moment and open an issue on testing-farm gitlab here: https://gitlab.com/testing-farm/general/-/issues/44

PR #966, opened 2022-06-29
Migrate to Bootstrap 5 / Fedora Bootstrap 2

ryanlerch:
This migrates the base templates, and the FAS, Centos, and Default
themes to use Bootstrap 5 / Fedora Bootstrap 2 (which uses Bootstrap 5)

A few things have changed between majopr versions of boostrap, but the
main things we had to migrate here were:

* left and right margin / padding / text alignment utilities now use the
  start and end terminology.
* data-* attributes that bootstrap 4 used to enable javascript things
  now use the data-bs-* prefix
* form-groups are no more -- we just use the padding utilities for these
  now
* finally, bootstrap 5 doens't need jquery anymore. We still have to
  include it as we use it elsewhere, but ideally we migrate that code in
the future so we arent using jquery anymore.


abompard (changes requested):
Looks good! There's one thing missing though, there are still places that use Bootstrap 4's media object, (classes media & media-body), they must be replaced by some flex controls: https://getbootstrap.com/docs/5.1/utilities/flex/#media-object

ryanlerch:
Updated the media object stuff to just flex utils as suggested -- this makes the lists look normal again!

PR #1153, opened 2023-03-02
Use poetry-core

jonathanspw:
poetry-core is all that is needed. This loosens build requirements so builds are possible against EPEL9 where `poetry-core` is package but `poetry` is not.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -191,5 +191,5 @@ testpaths = [
 ]
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
+requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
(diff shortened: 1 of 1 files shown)


abompard (approved):

pyproject.toml
```diff
@@ -191,5 +191,5 @@ testpaths = [
 ]
 
 [build-system]
-requires = ["poetry>=1.0.0"]
-build-backend = "poetry.masonry.api"
+requires = ["poetry-core>=1.0.0"]
```
hroncok: The version constraint is redundant. There has never been anything older.

```suggestion
requires = ["poetry-core"]
```

hroncok (approved):

PR #1287, opened 2023-06-30
Fix the boolean values not showing up in the profile page

### LINKED ISSUE #1202: "Private" checkbox on profile page not working.
The private checkbox on the edit profile page doesn't work, when saving and not modifying anything else it gives the error "no modifications to be performed". When saving and modifying something else it does appear to be saved, but when going back to the profile and going back to the edit profile page it just becomes unchecked again.

abompard:
Fix: #1202

### DIFF
--- a/news/1202.bug
+++ b/news/1202.bug
@@ -1,0 +1,1 @@
+Fix the boolean values (checkboxes) not showing up in the profile page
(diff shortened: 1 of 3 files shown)


davidkirwan (approved):
+1

PR #1352, opened 2024-01-18
Update the Flask version requirement to v3.0.0

### LINKED ISSUE #1350: flask 3.0.0 support?
Fedora rawhide has moved on to flask 3 and noggin package there doesn't build anymore: 

https://bugzilla.redhat.com/show_bug.cgi?id=2257105

we have: 

flask = "^2.0.0"

Can we enable 3.0.0 support and run tests to confirm it's good?

gridhead:
Update the dependency lockfile accordingly

Fixes #1350 

Please see https://github.com/fedora-infra/noggin/issues/1350#issuecomment-1897972752 for the reference on the tests run in the local development environment.


abompard:
Sorry but I don't think this is actually necessary for Flask support. The fact that it does not change the `pyproject.toml` file should have given me enough of a hint...
We already support Flask 3.x, the RPM doesn't see that because it wasn't at the latest released version.

PR #1353, opened 2024-01-23
Relocate the 'Leave Group' button for better visibility

### LINKED ISSUE #600: The "Leave group" button is too hidden, when there are many sponsors
The `Leave group` button is attached to the `Members` section. However, if there are a lot of sponsors, the button is not visible on the screen at all. And if people don't know where to look for it, they might not find it. As an artificial example, here's the Fedora's `packager` group:
https://accounts.fedoraproject.org/group/packager/
You have to scroll about 5 pages down (on my screen) to about middle of the webpage, where you can finally find a very subtle `Leave group` button.
It would make more sense to have the button at the very top, together with group description etc.

gridhead:
Fixes #600

### DIFF
--- a/noggin/templates/group.html
+++ b/noggin/templates/group.html
@@ -10,7 +10,7 @@
   <div class="bg-light border-bottom">
   <div class="container section">
     <div class="row py-4" data-section="identity">
-      <div class="col">
+      <div class="col col-9">
           <h3 class="mb-0">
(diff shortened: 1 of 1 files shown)


gridhead:
The button placement should be responsive across widths on various popular devices.

image

image

image

ryanlerch (approved):

PR #1389, opened 2024-03-27
Fix noggin-aaa package name in installation docs

eliba:
Correct Python package name should be `noggin-aaa`: https://pypi.org/project/noggin-aaa/

From https://pypi.org/project/noggin/:
> noggin is a simple tool for logging and plotting measurements during an experiment. [...]


gridhead (approved):
LGTM.

abompard (approved):

Issue #1409, opened 2024-05-27
Add more lines for blogs and rss feeds


phsmoura:
People in fedora planet community asked to add more lines so they can add different blogs.

abompard:
Fixed by #1410

PR #1410, opened 2024-05-27
Add more lines for blogs and rss feeds. Fixes: 1409

phsmoura:
Just changing here at noggin seems free-ipa and fasjson doesnt need to be updated, but let me know if it does or if this PR needs adjustments

related issue: #1409

### DIFF
--- a/noggin/form/edit_user.py
+++ b/noggin/form/edit_user.py
@@ -80,6 +80,11 @@ class ProtocolAndNickField(TypeAndStringField):
             raise ValidationError(_("This does not look like a valid server name."))
 
 
+def https_required(form, field):
+    if not field.data.startswith('https://'):
+        raise ValidationError('URL should start with "https://".')
+
+
 class UserSettingsProfileForm(BaseForm):
     firstname = StringField(
         _('First Name'),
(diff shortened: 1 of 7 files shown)


abompard:
OK I made quite a few commits, one of them is adjusting to the fact that `website_url` is not set as multi-valued in FAS (see https://github.com/fedora-infra/freeipa-fas/issues/151). You may want to fix this one before going through with this one.

nirik:
Would it be possible to enforce https here too?

phsmoura:
> Would it be possible to enforce https here too?

That would help users because planet drops HTTP links. 
@abompard : Do you know how we could block HTTP inputs?

abompard:
Sure, WTForms supports validators, there may be one for HTTPS urls in the URLFields, and if not there's always the regex validator

phsmoura:
> Sure, WTForms supports validators, there may be one for HTTPS urls in the URLFields, and if not there's always the regex validator

Seems WTForms doesnt have a validator to check HTTPS, so I created a new class for it and wrote a few unit tests.
I wasnt able to see it working tho, had trouble with the VMs using vagrant. Will try again on Monday.

Let me know if it needs more changes, maybe this PR is still a WIP

phsmoura:
@abompard / @nirik : What if we keep this issue to make these fields multivalued and open a new one requesting to enforce HTTPS in these fields? That way, we could keep the tasks within the scope of the issue.

phsmoura:
It's now blocking any link that doesn't start with HTTPS, but I'm still having trouble adding more lines to the `website` and `RSS` fields. I'll keep trying to do this, but any help is very appreciated.

phsmoura:
It works now. Its possible to add multiple lines for Blog and RSS fields and it only allows HTTPS, but still need to add unit tests for  `https_required`. Ill do it tomorrow

noggin/form/edit_user.py
```diff
@@ -80,6 +81,10 @@ def _validate(form, field):
 
 
 class UserSettingsProfileForm(BaseForm):
+    def _https_required(form, field):
```
abompard: Why make it a member of the class? This construct is a bit unusual, it should either be:
1. outside the class, as before
2. a method, but then it should have `self` as a first argument and called on the instance, which is not what we want for a validator
3. a class method, but then it should be decorated with `@classmethod` and take `cls` as the first argument
4. a static method, but then it should be decorated with `@staticmethod`

I think keeping it outside (option 1) is best here, so it can possibly be reused by another form later.

abompard (changes requested):

noggin/form/edit_user.py
phsmoura: Sorry, I thought it'd be used only for that case, but if it can possibly be reused I agree with option 1. Just did the changes.

abompard (approved):
Looks good, thanks!

PR #1422, opened 2024-06-21
Emit a message at the event of removing user from a group

### LINKED ISSUE #1415: Send a message at the event of removing a user from a group.
Message is sent when user is added to a group, but not when removed.
This functionality would be needed for actions that need to happen when a user is removed.

LenkaSeg:
Part of the effort of when a user is removed from packager group, user is also removed from all packager-related group, original issue here: https://pagure.io/fedora-infra/toddlers/issue/190

This depends on [PR #372](https://github.com/fedora-infra/noggin-messages/pull/372) in noggin-messages, that has to land first.

To test it locally, checkout the aforementioned PR from noggin-messages and add this in pyproject (correct the relative path to match your setting):

```
diff --git a/pyproject.toml b/pyproject.toml
index 52f04fa..40fbe23 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,8 @@ Flask-Mail = "^0.9.1 || ^0.10.0"
 gunicorn = {version = "^20.0 || ^21.0.0 || ^22.0.0", optional = true}
 fedora-messaging = ">=2.1.0, <4.0"
 backoff = ">=1.6.0, <3.0"
-noggin-messages = "^1.0.0"
+#noggin-messages = "^1.0.0"
+noggin-messages = {path = "../noggin-messages", develop = true}
 whitenoise = ">=5.3, <7.0"
 flask-babel = "^4.0.0"
 flask-healthz = "^1.0.0"
```
In my environment all the tests pass

image


LenkaSeg:
Tests are failing on the poetry.lock file, which is now generated with using the local modified noggin-messages dependency, so... not sure something can be done until noggin-messages PR gets solved.

noggin/controller/group.py
```diff
@@ -10,7 +10,7 @@
 from noggin.utility.controllers import group_or_404, with_ipa
 from noggin.utility.pagination import paginated_find
 from noggin.utility.templates import undo_button
-from noggin_messages import MemberSponsorV1
+from noggin_messages import MemberRemovedV1, MemberSponsorV1
```
abompard, line 13: Because of this, you need a minimal version on `noggin-messages` in `pyproject.toml` (`^1.1.0`). Updating the lockfile is not enough.

abompard (changes requested):
Also, please add a news fragment.

noggin/controller/group.py
LenkaSeg: Ah, yes, totally missed that.

abompard (approved):
Looks good, thanks!

PR #1450, opened 2024-09-27
Fix :  the Search Text field blocking by Toast

### LINKED ISSUE #1430: “Welcome” toast blocks access to search text field
After logging into accounts.fedoraproject.org and the “Welcome, …!” dialog is confirmed or fades out by itself, it still blocks the search text field underneath.

Apparently, the toast CSS from fedora-bootstrap tries to set `display: none` but a `d.flex > flex` rule wins by specificity:

Screenshot from 2024-07-12 12-26-49

Sachin2095:
## What does this PR do? 
In the User Profile Page after login , The Welcome Toast was blocking the Search Text field . 

Fixes #1430 
fixes_1430

## Type of change
-Bug Fix(non-Breaking change which fixes are  issue)

## How should this be tested ?
-[] After login [here](https://fedoraproject.org/) go to the user profile . 
-[] Click on  Search text field , This must be blocked by the toast . 
-[] go to inspect of that page and find out of the div of toast container and set (mt-0 ) then , issue will be resolved . 

## Mandatory Tasks 
-[X] Make sure you have self-reviewed the code . A decent size PR without self-review might be rejected .

### DIFF
--- a/noggin/themes/almalinux/templates/main.html
+++ b/noggin/themes/almalinux/templates/main.html
@@ -43,7 +43,7 @@
     </nav>
     {% with flashes = get_flashed_messages(with_categories=True) %}
         {% if flashes %}
-        <div class="container flash-messages mt-5">
+        <div class="container flash-messages mt-0">
             {% for category, flash in flashes %}
             <div class="alert alert-{{ category }} alert-dismissible fade show">
                 {{ flash }}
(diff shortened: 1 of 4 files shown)


abompard:
Thanks, this has already been fixed in #1455.

PR #1492, opened 2025-03-07
Add a step to sign agreements on registration

### LINKED ISSUE #1488: Consider adding agreements to new user flow
From: https://pagure.io/fedora-infrastructure/issue/12429 downstream

"note in advance: I am not the most technical person so if I use the wrong terminology - apologies but bear with me here

It would be nice that when you create your FAS account (or Fedora stage account) to have the user agreement as part of the registration process instead of going to your profile and the look for it to accept it there - after the registration.
Suggestion: Have it as part of the registration process (accept/Decline) either as a pop up or separate page in the registration process
This would greatly improve the user experience and ensures it won't be forgotten. This may be helpful espec new Fedora users and/or less technical experienced users.

When do you need this to be done by? No ETA - this is a UX improvement suggestion, low priority"

I'll add that this could have a 'look at agreements now' or 'skip' type thing if someone didn't want to do them right at registration.

abompard:
Fixes: #1488

This adds a step to sign agreements at the end of the registration process.

Example:

image

I'm happy to change the wording as necessary!

### DIFF
--- a/noggin/controller/registration.py
+++ b/noggin/controller/registration.py
@@ -20,12 +20,14 @@ from translitcodec import codecs
 from unidecode import unidecode
 
 from noggin.app import csrf, ipa_admin, mailer
+from noggin.form.edit_user import UserSettingsAgreementSign
 from noggin.form.register_user import (
     PasswordSetForm,
     RegisteringActionForm,
     ResendValidationEmailForm,
 )
 from noggin.l10n import guess_locale
+from noggin.representation.agreement import Agreement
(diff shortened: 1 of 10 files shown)


gridhead:
One minor nitpick: If the signing is optional, could you please reword the `Continue` button to either `Skip signing` or `Continue without signing`? Everything else looks good to me after skimming.

abompard:
Good idea, here's what it looks like now.

Before signing:
image

After signing:
image

PR #1525, opened 2025-08-20
Add Additional Email and use a Checkbox to Toggle Display of Fedora Project Email Alias

sgollor:
This pull request is for a feature that adds a check box in the user settings email to display the Fedora Project email alias.

When checked, the @fedoraproject.org email address is shown in the user profile and avatar/Gravatar, and when unchecked, the primary email is shown.

I did not add a new attribute in IPA. Instead, the user's choice is stored as a tag in the description attribute in IPA.

To turn on = ipa.user_mod(user.username, addattr=[f"description={description_tag}"])
Turn off = ipa.user_mod(user.username, delattr=[f"description={description_tag}"])

The user is also able to add and remove additional email addresses.

If the box is checked but the user has no @fedoraproject.org email in their list, we show a small warning sign, and nothing is changed.

### DIFF
--- a/noggin/controller/user.py
+++ b/noggin/controller/user.py
@@ -109,7 +109,7 @@ def _user_mod(ipa, form, user, details, redirect_to):
                 raise FormError("non_field_errors", e.message)
         flash(
             Markup(  # nosec B704
-                f'Profile Updated: <a href="{url_for(".user", username=user.username)}">'
+                f'Profile Updated: <a href=\"{url_for(".user", username=user.username)}\">'
                 'view your profile</a>'
             ),
             'success',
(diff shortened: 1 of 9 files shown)


abompard:
This PR seems a little too complex for what it needs to do.
Also, I'm not a fan of hijacking the description field to store this boolean.
And for now, we don't want to allow users to add more email addresses. Mostly because there's no way of sorting multivalued elements in LDAP and we'd need a primary address out of the emails.
All users that are part of the `fedora-contributor` group have a fedoraproject.org email alias, so you can use that for the warning.

abompard:
One more thing: if you're using an LLM to help you with the code, please add the tag:
```
Assisted-By: <name of the LLM>
```
to the git commits.

PR #1557, opened 2025-12-06
Add FPCA warning to agreement signing page. Fixes: #1530

### LINKED ISSUE #1530: not signing FPCA should warn that it will prevent you from logging in
## Status Quo 

Image

context: https://old.reddit.com/r/Fedora/comments/1n9pwon/how_do_i_login_to_idfedoraprojectorg/

Despite being marked as "Optional", not signing Fedora Project Contributor Agreement (FPCA) will prevent user from logging in to almost entirety of fedora ecosystem such as https://pagure.io and https://discussion.fedoraproject.org/.

## Suggestion

either:
- add an warning that not signing FPCA will prevent you from logging in to most fedora 
- on id.fedoraproject.org, when user hasn't signed to FPCA, return error message `You must sign FPCA` instead of generic `failed to authenticate` error message. I think this would be the best solution, but not sure whether this repo is the correct place to open issue.

Jkhall81:
Add FPCA warning to agreement signing page.

Fixes: #1530


Jkhall81:
Hi! I'm new to contributing to noggin. 

This PR adds a warning to the agreement signing page when users haven't signed the FPCA, addressing issue #1530. The warning explains that not signing will prevent login to other Fedora services.

I placed the warning in `registration-agreements.html` since that's the page shown in the issue screenshot. Please let me know if:
1. The warning text needs adjustment
2. It should also go somewhere else (like during authentication)
3. There are any other changes needed

Thanks

noggin/templates/registration-agreements.html
```diff
@@ -13,6 +12,25 @@
...
+                {% endif %}
+              {% endfor %}
+              
+              {% if fpca and fpca.name not in user.agreements %}
+                <div class="alert alert-warning my-3">
+                  <strong>{{ _("You have not signed the Fedora Project Contributor Agreement (FPCA).") }}</strong><br>
+                  {{ _("Not signing the FPCA will prevent you from logging in to many Fedora services such as Pagure, Fedora Discussion, and other contributor platforms.") }}
+                </div>
+              {% endif %}
+              <!-- END FPCA WARNING -->
```
abompard: You can move this block in the `_agreements_form.html` file, where there is already a loop over the agreements.

This way you don't need the block above where you set the `fpca` variable.

abompard (changes requested):
Thanks for the PR, I have a suggestion.

noggin/templates/registration-agreements.html
Jkhall81: Thanks for the feedback! I moved the FPCA warning block into _agreements_form.html as suggested and removed the logic from registration-agreements.html.

I also tried setting up a full Noggin development environment so I could walk through the UI changes directly, but I wasn’t able to get the user agreement flow working in my test setup. I’ll keep improving my local setup, but in the meantime I relied on reading the templates and following your guidance.

Please let me know if anything else should be adjusted!

noggin/templates/_agreements_form.html
```diff
@@ -1,6 +1,21 @@
 <div class="list-group">
     {% for agreement in agreementslist %}
-    <div class="list-group-item">
+
+      {# Determine whether this agreement is the FPCA #}
+      {% set is_fpca = agreement.slug == "fpca"
+                        or agreement.name|lower == "fpca"
+                        or agreement.name|lower.startswith("fedora project contributor agreement") %}
```
abompard: This doesn't work: 
`jinja2.exceptions.TemplateAssertionError: No filter named 'lower.startswith'.`

abompard (changes requested):
Thanks. I have one more comment and a suggestion: instead of hardcoding the name ("fpca" or "fedora project contributor agreement"), I think we should have a configuration value named, for example, `AGREEMENT_WARNINGS` that would be a dict where keys are agreement names and values are the warning message that should be displayed when the corresponding agreement is not signed.
This way we could make Noggin more independant from Fedora and it could be more easily reused by other projects.

noggin/defaults.py
```diff
@@ -61,3 +61,11 @@
 
 # Cheat code to toggle Fedora Messaging support
 FEDORA_MESSAGING_ENABLED = False
+
+# Agreement warnings shown when the corresponding agreement is not signed.
+# Keys must match the agreement.name from FreeIPA.
+AGREEMENT_WARNINGS = {
+    "Fedora Project Contributor Agreement": _(
```
abompard, line 70: For that to work you need to add `from flask_babel import _` at the top of the file.

abompard (changes requested):
Very nice, thanks!

For some ~stup~historic reason, the name of the FPCA agreement in the devel environment is `FPCA`, not `Fedora Project Contributor Agreement`. Could you add the same config block (with the `FPCA` key) to the `devel/ansible/roles/noggin/templates/noggin.cfg` config file as well please?

Also, please add a changelog fragment as described [in this doc](https://noggin-aaa.readthedocs.io/en/latest/contributing.html#release-notes).

Almost there! :-)

noggin/controller/registration.py
```diff
@@ -355,6 +355,7 @@ def sign_agreements(ipa):
         user=user,
         agreementslist=agreements,
         all_signed=all(agreement.name in user.agreements for agreement in agreements),
+        agreement_warnings=current_app.config.get('AGREEMENT_WARNINGS', {}),
```
abompard: Since you've added the key to `defaults.py`, you know it's there, you can use:
```suggestion
        agreement_warnings=current_app.config["AGREEMENT_WARNINGS"],
```

noggin/defaults.py
```diff
@@ -61,3 +61,11 @@
...
+        "Not signing the FPCA will prevent you from logging in to many Fedora services such as Pagure, Fedora Discussion, and other contributor platforms."
```
abompard: This is flagged as too long a line by `tox -e lint`.

noggin/templates/registration-agreements.html
```diff
@@ -13,6 +12,7 @@
         <div class="card">
             <div class="card-body">
               <h4 class="card-title">{{_("Optional agreement signing")}}</h4>
+
```
abompard: This line has empty spaces.

devel/ansible/roles/noggin/templates/noggin.cfg
```diff
@@ -67,3 +67,11 @@ MAIL_SUPPRESS_SEND = True
...
 BASSET_URL = None
 SPAMCHECK_TOKEN_EXPIRATION = 60  # in minutes
+
+# Agreement warnings shown when the corresponding agreement is not signed.
+# Keys must match the agreement name from FreeIPA.
+AGREEMENT_WARNINGS = {
+    "FPCA": "Not signing the FPCA will prevent you from logging in to many Fedora "
+            "services such as Pagure, Fedora Discussion, and other contributor "
+            "platforms."
+}
```
abompard, line 81: This file will need the `from flask_babel import _` as well.

noggin/defaults.py
```diff
@@ -61,3 +62,13 @@
 
 # Cheat code to toggle Fedora Messaging support
 FEDORA_MESSAGING_ENABLED = False
+
+# Agreement warnings shown when the corresponding agreement is not signed.
+# Keys must match the agreement.name from FreeIPA.
+AGREEMENT_WARNINGS = {
+    "FPCA": _(
```
abompard: The default agreement name is `Fedora Project Contributor Agreement`.

noggin/defaults.py
```diff
@@ -1,5 +1,6 @@
 # This file contains the default configuration values
 import socket
+from flask_babel import _
```
abompard, line 4: Black suggests adding a space before this line (you can check with `tox -e format`).

abompard (changes requested):
Just a few fixes left.
One slightly bigger thing: the `_agreements_form.html` template is also used by the `noggin.controller.user.user_settings_agreements()` view. As a result we get an error because the `agreement_warnings` template variable is not defined in this view. I see 2 ways of solving this:
- we define the variable in the `user_settings_agreements()` view as well
- instead of defining a variable in the `sign_agreements()` view, we use the fact that the app config is available in the templates directly in the `config` variable, and change `agreement_warnings` to `config["AGREEMENT_WARNINGS"]` in `_agreements_form.html`.

I would prefer the second option because it's less duplicated code.

You can check that is all works by running the unit tests with `tox -e py312-unittest`

Jkhall81:
Thanks for the clarification on the devel environment agreement name — I had misread that part earlier.

I’ve updated the configuration accordingly. I also set up tox and poetry locally so I could run the full test suite, and everything now passes.

`tox -e format` and `tox -e lint` are clean as well.

Let me know if anything else needs adjustment!

abompard:
Hey, thanks for the changes! My last comment is still relevant, about changing `agreement_warnings` to `config["AGREEMENT_WARNINGS"]` in `_agreements_form.html`.

Also, I've made a small fix to the translations in one of the files you touched, please rebase.

Jkhall81:
I rebased..  there was only a change in `registration-agreements.html`.

Also in `_agreements_form.html` I had already updated this line:

```jinja
{% set warning = config["AGREEMENT_WARNINGS"].get(agreement.name) %}
````

This is the one you are referring to correct?

File: news/1530.feature [hunk 10]
```diff
@@ -0,0 +1 @@
+Add warning to agreements page explaining that not signing the FPCA prevents logging in to most Fedora services
```
abompard: This file seems to have line endings in dos format (`\r\n`), please convert it to unix format (`\n`).

noggin/controller/registration.py
```diff
@@ -355,6 +355,7 @@ def sign_agreements(ipa):
         user=user,
         agreementslist=agreements,
         all_signed=all(agreement.name in user.agreements for agreement in agreements),
+        agreement_warnings=current_app.config["AGREEMENT_WARNINGS"],
```
abompard: I don't think you need this anymore, since you're now using the `config` template variable.

noggin/templates/registration-agreements.html
```diff
@@ -13,12 +12,9 @@
...
               <h4 class="card-title">{{_("Optional agreement signing")}}</h4>
-              <p>{{ngettext(
-                "Hello %(name)s. Please sign the agreement below if you agree with it.",
-                "Hello %(name)s. Please sign the agreements below if you agree with them.",
-                agreementslist|length,
-                name=user.name
-              )}}</p>
+
+              <p>{{_("Hello %(name)s. Please sign the agreement(s) below if you agree with them.", name=user.name)}}</p>
+
```
abompard, line 24: Hmm no, please leave the plural-aware translation string in place.

abompard (changes requested):
Still a few fixes, thanks!

abompard:
OK, looks good! Please rebase on the current develop branch, there's been an update to a couple translated strings that will conflict, but it should be easy to merge since you're only changing quotes and newlines.

Jkhall81:
Rebase completed! I pulled in the latest develop changes, resolved the template string conflicts, and force-pushed the updated branch. Let me know if anything else needs adjusting.

abompard (approved):
Looks good, thanks a lot for your contribution!

Jkhall81:
Very cool.

PR #1559, opened 2025-12-10
Avoid confusion that users have to sign an agreement that they could not yet review

py0xc3:
This PR solves this Discourse discussion: https://discussion.fedoraproject.org/t/signing-up-to-fedora-deters-users-click-sign-button-before-agreement-can-be-read-reviewed/176370

I have little experience with this type of implementation: it needs further review. Especially the line 9 (`{% if agreement.name in user.agreements %}`) of `.../noggin/templates/_agreements_form.html` makes me wonder if the button name "Review" of line 12 will fit all circumstances in which the line might be called. I also presume the buttons are autogenerated from the given names.

Based on the issue elaborated in the Discourse topic, this PR shall do just two things (with the screenshot of the Discourse topic in mind): 
1) change the "sign" button the user has to click before reviewing the actual content they have to sign (to avoid the confusion) to the name "review".
2) Adjust the text presented to the user in order to fit the button & the need to first review and then sign in the next step (after reviewing).

Hope that safes some time :)

### DIFF
--- a/noggin/templates/_agreements_form.html
+++ b/noggin/templates/_agreements_form.html
@@ -9,7 +9,7 @@
             {% if agreement.name in user.agreements %}
             <a data-bs-toggle="modal" data-bs-target="#agreement-modal-{{agreement.slug}}" class="btn btn-link" href="javascript:void(0)">{{ _("View agreement") }}</a>
             {% else %}
-            <a data-bs-toggle="modal" data-bs-target="#agreement-modal-{{agreement.slug}}" class="btn btn-primary" href="javascript:void(0)">{{ _("Sign") }}</a>
+            <a data-bs-toggle="modal" data-bs-target="#agreement-modal-{{agreement.slug}}" class="btn btn-primary" href="javascript:void(0)">{{ _("Review") }}</a>
             {% endif %}
           </div>
         </div>
--- a/noggin/templates/registration-agreements.html
+++ b/noggin/templates/registration-agreements.html
@@ -13,7 +13,12 @@
         <div class="card">
             <div class="card-body">
               <h4 class="card-title">{{_("Optional agreement signing")}}</h4>
-              <p>{{_("Hello %(name)s. Please review the agreement(s) below and then sign them if you agree with their content.", name=user.name)}}</p>
+              <p>{{ngettext(
+                "Hello %(name)s. Please review the agreement below and then sign it if you agree with its content.",
+                "Hello %(name)s. Please review the agreements below and then sign them if you agree with their content.",
+                agreementslist|length,
+                name=user.name
+              )}}</p>
               {% include '_agreements_form.html' %}
             </div>
             <div class="card-footer d-flex justify-content-between">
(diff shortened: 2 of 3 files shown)


abompard (changes requested):
Thanks! Please update the translated strings as well, following [the contributors documentation](https://noggin-aaa.readthedocs.io/en/latest/contributing.html#translations).

This command should do it:
`poetry run pybabel extract -F babel.cfg -o noggin/translations/messages.pot noggin`

You'll probably need to rebase before running it because I've made a few changes to some translations and they will probably conflict.

py0xc3:
Project is now rebased. But when I try your poetry cmd, I end up with an error message:
```
[py0xc3@toolbx noggin]$ poetry run pybabel extract -F babel.cfg -o noggin/translations/messages.pot noggin

The Poetry configuration is invalid:
  - The fields ['authors', 'description', 'name', 'version'] are required in package mode.
```

I run it from within the repo's main directory, in which the file `babel.cfg` is located; `poetry` recognizes that this is a poetry project and when testing if that is a general problem on my system, `poetry` outputs `Poetry could not find a pyproject.toml file in <dir> or its parents` if I run it outside of the project, so the error looks to be specific to the project and not poetry configuration in general, though I am not 100% sure about that (?)

I am not sure if it is intended that I update the file `pyproject.toml` or if that error message has a different background? (Never worked with `poetry` before and don't find much useful around this specific error output, except an error ticket indicating this to be an error linked to the content of the project's `pyproject.toml`)

abompard:
> [py0xc3@toolbx noggin]$ poetry run pybabel extract -F babel.cfg -o noggin/translations/messages.pot noggin
> 
> The Poetry configuration is invalid:
>   - The fields ['authors', 'description', 'name', 'version'] are required in package mode.

This is probably because you're using version 1.x of Poetry, which does not know how to handle the now standard `[project]` key in `pyproject.toml`. Is there a way for you to upgrade to Poetry 2.x?

abompard:
If not, it's not a big problem, I can run the command in the further commit.

py0xc3:
> This is probably because you're using version 1.x of Poetry,

True, thanks for the hint! I was doing this in a toolbox with a f42 container, while f42 is still on 1.x poetry. I install it natively on my Fedora, which is f43 & therefore has poetry 2.x in its repos. I'll do this later and update here once done.

py0xc3:
I successfully ran `poetry run pybabel extract -F babel.cfg -o noggin/translations/messages.pot noggin` & pushed it to the PR, I also already rebased to your recent commit :) Let me know if more is to be done :+1:

py0xc3:
I haven't worked with github code committing for years. It seems to not recognize my local sign-off, and now it just gets ugly with subsequent commits that don't solve the problem. I need to check this out more deeply before keeping to commit commits github doesn't recognize as signed off. But I cannot invest more time in this now, not sure if you can accept the PR that way. I will have to check out how to resolve this later or in the next days. If I cannot find much time I might just do the quick & ugly solution, remove this PR and the repo, and create a new fork that just has one online commit that leads to the final result. Sorry ^^

py0xc3:
I created a new PR from a clean branch. The changes are equal to this one, but not the mess with commits and the sign-off issue. I close this one. The new one is #1564

PR #1564, opened 2025-12-11
Avoid confusion that users have to sign an agreement that they could not yet review (PR from clean branch)

py0xc3:
This replaces the earlier PR #1559 with a clean branch but the same changes.

---------

Original PR text:

This PR solves this Discourse discussion: https://discussion.fedoraproject.org/t/signing-up-to-fedora-deters-users-click-sign-button-before-agreement-can-be-read-reviewed/176370

I have little experience with this type of implementation: it needs further review. Especially the line 9 (`{% if agreement.name in user.agreements %}`) of `.../noggin/templates/_agreements_form.html` makes me wonder if the button name "Review" of line 12 will fit all circumstances in which the line might be called. I also presume the buttons are autogenerated from the given names.

Based on the issue elaborated in the Discourse topic, this PR shall do just two things (with the screenshot of the Discourse topic in mind):

    change the "sign" button the user has to click before reviewing the actual content they have to sign (to avoid the confusion) to the name "review".
    Adjust the text presented to the user in order to fit the button & the need to first review and then sign in the next step (after reviewing).

Hope that safes some time :)

### DIFF
--- a/noggin/templates/_agreements_form.html
+++ b/noggin/templates/_agreements_form.html
@@ -9,7 +9,7 @@
             {% if agreement.name in user.agreements %}
             <a data-bs-toggle="modal" data-bs-target="#agreement-modal-{{agreement.slug}}" class="btn btn-link" href="javascript:void(0)">{{ _("View agreement") }}</a>
             {% else %}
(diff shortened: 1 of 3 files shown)


py0xc3:
I think this one is ready to merge :) Let me know if further changes are necessary. Thanks for taking care ;)

abompard (approved):
Looks good, thanks!

PR #1593, opened 2026-06-10
fix(templates.py): fix the traceback when providing a https:// link

### LINKED ISSUE #1592: Traceback when IPA chat room field contains a https:// URL
When noggin tries to parse the IRC network and channel from IPA and it contains a `https://` link, it tracebacks:

```
[2026-06-10 17:59:18,314] ERROR in app: Exception on /group/flock-team/ [GET]
Traceback (most recent call last):
  File "/opt/app-root/src/noggin/utility/templates.py", line 60, in format_chat
    default_server = current_app.config["CHAT_NETWORKS"][scheme]["default_server"]
KeyError: 'https'
```

smoliicek:
this fixes https://github.com/fedora-infra/noggin/issues/1592

i currently just escape the url that we get from IPA, since the group info isn't something a user can change without going through the IPA UI - let me know if this should be done differently

### DIFF
--- a/noggin/utility/templates.py
+++ b/noggin/utility/templates.py
@@ -55,8 +55,8 @@ def format_chat(value, isnick):
         name = url.fragment
     scheme = url.scheme
     if scheme in ("http", "https"):
-        value = escape(value)
-        return Markup(f"""<a href="{value}">{value}</a>""")  # nosec B704
+        label = escape(value)
+        return Markup(f"""<a href="{value}">{label}</a>""")  # nosec B704
     if not scheme:
         scheme = "irc"
     try:
--- a/noggin/utility/templates.py
+++ b/noggin/utility/templates.py
@@ -3,7 +3,7 @@ from urllib.parse import urlparse
 
 from flask import current_app
 from flask_babel import lazy_gettext as _
-from markupsafe import Markup
+from markupsafe import Markup, escape
 
 
 def gravatar(email, size):
@@ -54,6 +54,9 @@ def format_chat(value, isnick):
     elif not name and url.fragment:
         name = url.fragment
     scheme = url.scheme
+    if scheme in ("http", "https"):
+        value = escape(value)
+        return Markup(f"""<a href="{value}">{value}</a>""")  # nosec B704
     if not scheme:
         scheme = "irc"
     try:
--- a/tests/unit/utility/test_templates.py
+++ b/tests/unit/utility/test_templates.py
@@ -80,6 +80,12 @@ def test_format_nickname_no_matrixto_args(request_context):
         assert str(format_nickname(value)) == expected_html
 
 
+def test_format_nickname_web_url(request_context):
+    value = "https://example.com/user"
+    expected_html = f'<a href="{value}">{value}</a>'
+    assert str(format_nickname(value)) == expected_html
+
+
 def test_format_nickname_invalid(request_context):
     with pytest.raises(ValueError) as e:
         format_nickname("invalid:/username")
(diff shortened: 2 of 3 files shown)


noggin/utility/templates.py
```diff
@@ -54,6 +54,9 @@ def format_chat(value, isnick):
     elif not name and url.fragment:
         name = url.fragment
     scheme = url.scheme
+    if scheme in ("http", "https"):
+        value = escape(value)
+        return Markup(f"""<a href="{value}">{value}</a>""")  # nosec B704
```
james-antill: escape() uses entity escaping, for HTML data. Eg. < becomes &lt; etc.

I guess you want to use the unescaped value as the href, and the escaped value as the entity data of the a.

smoliicek: sounds fair, will change

tests/unit/utility/test_templates.py
```diff
@@ -80,6 +80,13 @@ def test_format_nickname_no_matrixto_args(request_context):
         assert str(format_nickname(value)) == expected_html
 
 
+def test_format_nickname_web_url(request_context):
```
abompard: You don't actually need the `request_context` fixture here because you never go to the part where `current_app` is used. But that's a detail 🙂

tests/unit/utility/test_templates.py
```diff
@@ -165,3 +172,11 @@ def test_format_nickname_invalid_with_config(app, request_context, mocker):
 def test_format_channel(request_context, value, expected):
     expected_html = '<a href="{href}" title="{title}">{name}</a>'.format(**expected)
     assert str(format_channel(value)) == expected_html
+
+
+@pytest.mark.parametrize("scheme", ["http", "https"])
+def test_format_channel_web_url(request_context, scheme):
```
abompard: Same here, you don't actually need `request_context`.

abompard (changes requested):
Thanks! A couple minor details, but please add a news fragment as well.

smoliicek:
okay, fixed all the things, should be ready for merging now!

noggin/utility/templates.py
```diff
@@ -54,6 +54,9 @@ def format_chat(value, isnick):
     elif not name and url.fragment:
         name = url.fragment
     scheme = url.scheme
+    if scheme in ("http", "https"):
+        label = escape(value)
+        return Markup(f"""<a href="{value}">{label}</a>""")  # nosec B704
```
abompard, line 59: Actually, the `href` attribute should also be escaped to prevent attribute injection attacks. If `value` contains quotes or other special characters, an attacker could break out of the href attribute.

Recommended fix:

```suggestion
        escaped_url = escape(value)
        return Markup(f"""<a href="{escaped_url}">{escaped_url}</a>""")  # nosec B704
```

This ensures both the href and display text are properly escaped.

tests/unit/utility/test_templates.py
```diff
@@ -165,3 +172,11 @@ def test_format_nickname_invalid_with_config(app, request_context, mocker):
 def test_format_channel(request_context, value, expected):
     expected_html = '<a href="{href}" title="{title}">{name}</a>'.format(**expected)
     assert str(format_channel(value)) == expected_html
+
+
+@pytest.mark.parametrize("scheme", ["http", "https"])
+def test_format_channel_web_url(scheme):
+    value = f"{scheme}://example.com/channel?a=1&b=2"
+    escaped_value = f"{scheme}://example.com/channel?a=1&amp;b=2"
+    expected_html = f'<a href="{value}">{escaped_value}</a>'
```
abompard, line 181: Taking into account the escaping in href:

```suggestion
    expected_html = f'<a href="{escaped_value}">{escaped_value}</a>'
```

tests/unit/utility/test_templates.py
```diff
@@ -165,3 +172,11 @@ def test_format_nickname_invalid_with_config(app, request_context, mocker):
...
+    assert str(format_channel(value)) == expected_html
```
abompard, line 182: Consider adding a security test case that validates proper escaping of malicious input:

```python
def test_format_channel_web_url_xss_attempt():
    value = 'https://evil.com" onclick="alert(1)'
    result = str(format_channel(value))
    # Should not allow breaking out of href attribute
    assert 'onclick="alert(1)"' not in result
```

abompard (changes requested):
Thanks for the changes! There's however a security concern with `format_chat`. If `value` contains a quote character, an attacker could break out of the href attribute. For example:
- Input: `https://evil.com" onclick="alert('XSS')`
- Output: `<a href="https://evil.com" onclick="alert('XSS')">{escaped}</a>`

This would create a clickable link that executes JavaScript.

PR #1598, opened 2026-07-07
Add OTP recovery codes and admin-assisted OTP reset

abbra:
### Summary

Add HOTP-based recovery codes for OTP-enrolled users, so they can
regain access if they lose their authenticator device. Also add an
admin-assisted OTP reset flow that lets designated IPA role members
wipe a user's tokens and issue fresh recovery codes on their behalf.

#### What's included

- **Recovery code generation**: counter-based (HOTP) token with 10
  eight-digit codes, created via the user's OTP settings page
- **Recovery code regeneration**: delete-and-recreate with password
  re-authentication
- **Mandatory recovery codes** (opt-in via config): auto-generate
  codes when the user enrolls their first OTP token
- **Admin OTP reset**: role-gated route that deletes all tokens for
  a target user and creates new recovery codes, with Fedora Messaging
  notification
- **Recovery code display**: remaining count, low-count warning,
  download/copy/print support via shared Jinja2 macros
- **OTP controller extraction**: existing OTP routes moved from
  `user.py` to `user_otp.py` for maintainability

#### Configuration

Five new settings in `noggin/defaults.py`:
- `OTP_RECOVERY_CODE_COUNT` (default: 10)
- `OTP_RECOVERY_DESCRIPTION` (default: "Recovery codes")
- `OTP_RECOVERY_LOW_THRESHOLD` (default: 3)
- `OTP_REQUIRE_RECOVERY_CODES` (default: False)
- `OTP_ADMIN_RECOVERY_ROLE` (default: None — disables admin reset)

### Test plan

- [ ] Generate recovery codes after enrolling an OTP token
- [ ] Regenerate codes and verify old codes stop working
- [ ] Verify admin reset button only appears for role members
- [ ] Verify admin reset deletes all tokens and issues new codes
- [ ] Verify recovery code count decrements after each use
- [ ] Verify low-count warning at threshold

Generated by Claude Code

### DIFF
--- a/docs/design-otp-recovery-codes.rst
+++ b/docs/design-otp-recovery-codes.rst
@@ -1,0 +1,791 @@
+=====================
+OTP Recovery Codes
+=====================
+
+:Author: Alexander Bokovoy
+:Status: Implemented
+:Created: 2026-07-02
+
+.. contents:: Table of Contents
+   :local:
+   :depth: 2
+
+
+Problem Statement
+=================
+
+When a Noggin user enrolls an OTP token for two-factor authentication,
+losing the authenticator device (phone loss, factory reset, hardware
+failure) locks them out completely. FreeIPA enforces OTP once any token
+is active for the user — there is no fallback. The only recovery path
+today requires an IPA administrator to delete the lost token and
+provision a new one.
+
+This is unacceptable for a self-service portal. Users must have a
+recovery mechanism they can set up *before* lockout occurs, without
+requiring administrator intervention.
+
+
+Goals
+=====
+
+1. Allow users to generate a set of single-use recovery codes during OTP
+   enrollment.
+2. Any recovery code can be used in place of a TOTP code at any login
+   prompt (Noggin, SSH, FreeIPA WebUI, etc.).
+3. No changes to FreeIPA schema, ``ipa-otpd``, or 389-ds SLAPI plugins.
+4. Fit naturally into Noggin's existing OTP management UI.
+5. Recovery codes can be regenerated (invalidating all previous codes).
+
+
+Non-Goals
+=========
+
+- Out-of-band identity verification (email link, manager approval) for
+  locked-out users who never set up recovery codes.
+- Per-token HOTP authentication window overrides (requires FreeIPA
+  schema changes).
+
+
+Design
+======
+
+Overview
+--------
+
+Recovery codes are implemented as an **always-active HOTP token** owned
+by the user. Each recovery code is a pre-computed HOTP value for a
+sequential counter position. Because the token stays enabled, the
+FreeIPA OTP validation path accepts recovery codes alongside the user's
+primary TOTP token with no server-side changes.
+
+This is the same pattern used by Google, GitHub, and most services that
+offer backup codes for 2FA.
+
+
+How It Works in FreeIPA
+-----------------------
+
+FreeIPA's OTP validation happens inside the ``ipa-pwd-extop`` SLAPI
+plugin (``prepost.c:ipapwd_pre_bind_otp()``). When a user authenticates
+with ``password + OTP``:
+
+1. The plugin calls ``otp_token_find()`` which searches LDAP for all
+   **active** tokens owned by the user. The LDAP filter excludes
+   disabled tokens (``ipatokenDisabled=TRUE``) and tokens outside their
+   validity window.
+
+2. ``otp_token_validate_berval()`` iterates over all matching tokens and
+   tries to validate the OTP code from the end of the password string
+   against each token (TOTP or HOTP).
+
+3. For an HOTP token, validation checks the code against
+   ``counter + offset`` for offsets 0 through ``ipatokenHOTPauthWindow``
+   (default: 10). On success, the counter advances past the matched
+   position. The counter never goes backwards, so each code is
+   single-use.
+
+Because the recovery HOTP token is active and owned by the user, it
+participates in this validation loop transparently. No changes to the
+FreeIPA C code, Python plugins, or LDAP schema are needed.
+
+
+Identifying Recovery Tokens
+----------------------------
+
+FreeIPA has no ``ipatokenPurpose`` attribute. Recovery tokens are
+distinguished by convention:
+
+- **Description**: Set to ``"Recovery codes"`` (the literal string, not
+  localized in LDAP — localization is only in the Noggin UI).
+- **Token type**: ``hotp`` (as opposed to ``totp`` for the primary
+  token).
+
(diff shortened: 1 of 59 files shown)


abbra:
I finally cleared up my attempt to add recovery codes, with the help of Claude Code.
Please read the proposed design document for details. This code relies on existing HOTP tokens feature and basically using a customized HOTP token to handle recovery codes. As this is a normal HOTP token, its codes are accepted at any login point (kerberos, Ipsilon, Noggin). Users supposed to not use the token for normal logins and since it is HOTP, it doesn't expire values with time.

abbra:
OTP handling was split into a separate `user_otp.py` file because it became too large.

The translation changes are mostly due to new strings in the recovery codes' section but they also conflict with just new translations update on the develop branch, so there is new context and commits that could otherwise be ignored.

nirik:
This is awesome. Thanks for working on it.

noggin/templates/_recovery_codes.html
```diff
@@ -0,0 +1,54 @@
+{% macro codes_grid(recovery_codes) %}
+<div class="row justify-content-center my-3">
+  <div class="col-md-10">
+    <div class="row">
+      {% for code in recovery_codes %}
+        <div class="col-6 mb-2">
+          <code class="fs-5" data-role="recovery-code">{{ loop.index }}. {{ code }}</code>
+        </div>
+      {% endfor %}
+    </div>
```
abompard, line 10: The suggested change below would let people select the codes with the mouse without selecting the indexes, and would not display the indexes in the same color and font as the codes themselves:

```suggestion
    <ol class="d-flex flex-row flex-wrap">
      {% for code in recovery_codes %}
        <li class="col-6 mb-2">
          <code class="fs-5" data-role="recovery-code">{{ code }}</code>
        </li>
      {% endfor %}
    </ol>
```

noggin/templates/user-settings-otp.html
```diff
@@ -123,6 +157,76 @@ <h5 id="pageheading">{{ _("OTP Tokens") }}</h5>
 
 {{ otp_notice() if otp_notice is defined and tokens }}
 
+<div class="card-body mt-4" id="recovery-codes">
```
abompard, line 160: You can wrap this block in `{% if tokens %}...{% endif %}` to hide it when there are no OTP tokens yet.

docs/design-otp-recovery-codes.rst
```diff
@@ -0,0 +1,791 @@
...
+Add to ``noggin/defaults.py``::
+
+    # Number of recovery codes to generate. Must not exceed the FreeIPA
+    # ipatokenHOTPauthWindow (default: 10).
+    OTP_RECOVERY_CODE_COUNT = 10
+    # Description string used to identify recovery tokens in LDAP.
+    OTP_RECOVERY_DESCRIPTION = "Recovery codes"
+
+The code count is a configuration value rather than a hard-coded constant
+so deployments with a larger ``ipatokenHOTPauthWindow`` can increase it.
```
abompard, line 161: Is there a particular reason why `OTP_RECOVERY_DESCRIPTION` is a configuration option rather than a constant? Are you thinking of a situation where an admin would want to change it from the default?

docs/design-otp-recovery-codes.rst
```diff
@@ -0,0 +1,791 @@
...
+user is not fully locked out — they can still authenticate with
+``password + recovery_code``.
+
+**Regeneration edge case**: The regenerate flow deletes the old recovery
+token before creating a new one. If the recovery token is the user's
+*only* active token, the delete will fail with
+``"Can't delete last active token"``. Noggin must handle this error and
+tell the user to add a regular OTP token first. In practice this is
+unlikely — the user would have to delete all TOTP tokens while keeping
+the recovery token, which is a degenerate state.
```
abompard, line 477: I'm not sure we should allow a situation where the user only has the recovery codes active (except in the admin-assisted case). I think that Noggin should prevent the removal of the last TOTP token, otherwise the "recovery codes" functionally fall back to being HOTP tokens that are required for login. As a user, I think I would expect removing my last (T)OTP token to remove the recovery codes as well, because there is nothing left to recover.
And since we don't allow removal of the last (T)OTP token by policy, we should have Noggin handle the situation (in place of `ipa-otp-lasttoken` if it cannot handle differently TOTP and HOTP tokens).

docs/design-otp-recovery-codes.rst
```diff
@@ -0,0 +1,791 @@
...
+- **test_generate_recovery_codes_uses_sha256**: Verify SHA-256
+  algorithm is used.
+- **test_generate_recovery_codes_zero_count**: Zero count returns
+  empty list.
+- **test_generate_recovery_codes_negative_count**: Negative count
+  returns empty list.
+- **test_generate_recovery_codes_unique**: Verify all codes in a set
+  are unique.
+- **test_generate_recovery_codes_custom_count**: Custom count
+  generates correct number of codes.
```
abompard, line 650: This list has a high chance of becoming obsolete overtime. Please use docstrings in the tests themselves.

docs/design-otp-recovery-codes.rst
```diff
@@ -0,0 +1,791 @@
...
+-----------------
+
+Against a live FreeIPA instance:
+
+- Enroll a TOTP token and generate recovery codes.
+- Authenticate with ``password + recovery_code`` (using one of the 10
+  codes).
+- Verify the used code no longer works (counter advanced).
+- Verify remaining codes still work.
+- Regenerate codes and verify old codes no longer work.
```
abompard, line 662: I don't see these integration tests in the PR, but we don't have any yet (only a placeholder). I could write these tests if you think they are useful and you'd rather not write them.

docs/design-otp-recovery-codes.rst
```diff
@@ -0,0 +1,791 @@
...
+Mandatory recovery codes on first OTP enrollment
+--------------------------------------------------
+
+When ``OTP_REQUIRE_RECOVERY_CODES`` is ``True``, confirming the first
+OTP token automatically generates recovery codes and displays them in
+the codes modal. The user must acknowledge saving the codes before
+proceeding. This ensures every OTP user has recovery codes from the
+moment they enable two-factor authentication.
+
+When the flag is ``False`` (the default), a flash message with a link
```
abompard, line 711: That's a really nice workflow, I think we want it by default 🙂

abompard (changes requested):
Hey! Thanks a lot for working on this. I have a few comments, so here's a first pass 🙂

abbra:
@abompard thank you for the review! I'm planning to look at this code after first week of August.

